PR #33, opened 2021-11-28
Deposit cache validating

### LINKED ISSUE #32: Bot should clear cache if data in cache is older than current block
If the daemon with fresh cache looks at the delayed node, it could find matching keys

avsetsin:
closes #32


infloop (approved):
LGTM

PR #77, opened 2023-01-11
Feat/keys api

Eddort:


sample.env
```diff
@@ -36,3 +36,40 @@ KAFKA_BROKER_ADDRESS_1=localhost:9092
...
 # Make sure there are enough ETH on the balance to send a transaction to stop the protocol
 WALLET_PRIVATE_KEY=0x0000000000000000000000000000000000000000000000000000000000000000
+
+# KEYS API
+
+# App
+KEYS_API_PORT=3001
+
+# CORS
+KEYS_API_CORS_WHITELIST_REGEXP=^https?://(?:.+?\.)?(?:lido|testnet|mainnet)\.fi$
```
avsetsin, line 46: keys api should not be externally available here. let's leave it blank to avoid unnecessary questions

src/guardian/guardian.service.ts
```diff
@@ -202,16 +258,15 @@ export class GuardianService implements OnModuleInit {
...
+
     const filteredIntersections = await this.excludeEligibleIntersections(
-      intersections,
+      keysIntersections,
       blockData,
     );
     const isFilteredIntersectionsFound = filteredIntersections.length > 0;
 
-    this.collectIntersectionsMetrics(intersections, filteredIntersections);
+    this.collectIntersectionsMetrics(keysIntersections, filteredIntersections);
```
avsetsin, line 269: this is called in the module loop, so the metrics will be overwritten by the last processed module

src/guardian/guardian.service.ts
```diff
@@ -202,16 +258,15 @@ export class GuardianService implements OnModuleInit {
...
 
     if (blockData.isDepositsPaused) {
       this.logger.warn('Deposits are paused', { blockHash });
```
avsetsin, line 272: Need to specify for which module

src/guardian/guardian.service.ts
```diff
@@ -231,64 +286,21 @@ export class GuardianService implements OnModuleInit {
...
 
-    const nextSigningKeysSet = new Set(nextSigningKeys);
+    const unusedKeysSet = new Set(unusedKeys);
     const intersections = depositedEvents.events.filter(({ pubkey }) =>
-      nextSigningKeysSet.has(pubkey),
+      unusedKeysSet.has(pubkey),
     );
 
     if (intersections.length) {
       this.logger.warn('Already deposited keys found in the next Lido keys', {
```
avsetsin, line 300: need to have module info here

src/guardian/guardian.service.ts
```diff
@@ -231,64 +286,21 @@ export class GuardianService implements OnModuleInit {
    * @param blockData - collected data from the current block
    * @returns list of keys that were deposited earlier
    */
-  public getNextKeysIntersections(
-    blockData: BlockData,
-  ): VerifiedDepositEvent[] {
+  public getKeysIntersections(blockData: BlockData): VerifiedDepositEvent[] {
```
avsetsin, line 289: I think we should explicitly put in the name that the intersections are searched for one module. As in all other methods that have the same behavior

avsetsin: There is also a problem with the semantic of the argument. It's called `blockData`, but it contains data for one module + general data. Consider splitting the data into several arguments

src/guardian/guardian.service.ts
```diff
@@ -525,19 +544,10 @@ export class GuardianService implements OnModuleInit {
...
-    const operatorsKeysUnusedTotal = operatorsKeysUnused.length;
-
-    this.operatorsKeysCounter.set({ type: 'total' }, operatorsKeys.length);
-    this.operatorsKeysCounter.set({ type: 'used' }, operatorsKeysUsedTotal);
-    this.operatorsKeysCounter.set({ type: 'unused' }, operatorsKeysUnusedTotal);
-    this.operatorsKeysCounter.set({ type: 'next' }, nextSigningKeys.length);
+    const { unusedKeys } = blockData;
+
+    const operatorsKeysTotal = unusedKeys.length;
+    this.operatorsKeysCounter.set({ type: 'total' }, operatorsKeysTotal);
```
avsetsin, line 550: need to have module id here or calc sum

src/keys-api/interfaces/RegistryKey.ts
```diff
@@ -0,0 +1,22 @@
+export type RegistryKey = {
```
avsetsin, line 1: I guess it's better to use a common interface for a Key rather than the RegistryKey

src/keys-api/interfaces/SRKeyListWithModule.ts
```diff
@@ -0,0 +1,13 @@
+import type { RegistryKey } from './RegistryKey';
+import type { SRModule } from './SRModule';
+
+export type SRKeyListWithModule = {
+  /**
+   * Keys of staking router module
+   */
+  keys: Array<RegistryKey>;
```
avsetsin, line 8: Key, not RegistryKey

src/keys-api/keys-api.service.ts
```diff
@@ -0,0 +1,51 @@
+import { Injectable, LoggerService, Inject } from '@nestjs/common';
+import { FetchService } from '@lido-nestjs/fetch';
+import { AbortController } from 'node-abort-controller';
+import { FETCH_REQUEST_TIMEOUT } from './keys-api.constants';
+import { SRModuleKeysResponse, SRModuleListResponse } from './interfaces';
+import { WINSTON_MODULE_NEST_PROVIDER } from 'nest-winston';
+import { Configuration } from 'common/config';
+
+@Injectable()
+export class KeysApiService {
```
avsetsin, line 10: I think we should collect the Keys API version and send it to the message broker

avsetsin: And it's worth collecting chainId from Keys API and checking it against the chainId of the council daemon EL API providers

src/keys-api/keys-api.service.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  public async getModulesList() {
+    return await this.fetch<SRModuleListResponse>('/v1/modules');
+  }
+
+  public async getUnusedModuleKeys(stakingModuleId: number) {
+    return await this.fetch<SRModuleKeysResponse>(
+      `/v1/modules/${stakingModuleId}/keys?used=false`,
+    );
+  }
```
avsetsin, line 50: it's better to fetch data for all modules at once

avsetsin (changes requested):

src/wallet/wallet.service.ts
```diff
@@ -139,17 +152,20 @@ export class WalletService implements OnModuleInit {
...
+  public async signPauseData({
+    prefix,
+    blockNumber,
+    stakingModuleId,
+  }: SignPauseDataParams): Promise<Signature> {
     const encodedData = defaultAbiCoder.encode(
-      ['bytes32', 'uint256'],
-      [prefix, blockNumber],
+      ['bytes32', 'uint256', 'uint24'],
+      [prefix, blockNumber, stakingModuleId],
```
avsetsin, line 168: stakingModuleId  is `uint256` according to the current contract code
https://github.com/lidofinance/lido-dao/blob/feature/shapella-upgrade/contracts/0.8.9/DepositSecurityModule.sol#L307-L311

src/wallet/wallet.service.ts
```diff
@@ -114,23 +117,33 @@ export class WalletService implements OnModuleInit {
...
-      ['bytes32', 'bytes32', 'uint256', 'uint256', 'bytes32'],
-      [prefix, depositRoot, keysOpIndex, blockNumber, blockHash],
+      ['bytes32', 'uint256', 'bytes32', 'bytes32', 'uint24', 'uint256'],
+      [
+        prefix,
+        blockNumber,
+        blockHash,
+        depositRoot,
+        stakingModuleId,
+        keysOpIndex,
```
avsetsin, line 145: stakingModuleId  is `uint256`

https://github.com/lidofinance/lido-dao/blob/feature/shapella-upgrade/contracts/0.8.9/DepositSecurityModule.sol#L406-L413

src/staking-router/staking-router.service.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  public async getStakingModules() {
+    return await this.keysApiService.getModulesList();
+  }
+
+  public async getStakingModuleUnusedKeys({ id, nonce }: SRModule) {
+    if (!this.isNeedToUpdateState(id, nonce))
+      return this.getStakingRouterKeysCache(id);
+
+    const srResponse = await this.keysApiService.getUnusedModuleKeys(id);
```
avsetsin, line 24: do we handle errors somewhere when fetching data from the Keys API?

src/staking-router/staking-router.service.ts
```diff
@@ -0,0 +1,48 @@
...
+    @Inject(WINSTON_MODULE_NEST_PROVIDER) protected logger: LoggerService,
+    protected readonly config: Configuration,
+    protected readonly keysApiService: KeysApiService,
+  ) {}
+
+  public async getStakingModules() {
+    return await this.keysApiService.getModulesList();
+  }
+
+  public async getStakingModuleUnusedKeys({ id, nonce }: SRModule) {
```
avsetsin, line 20: don't we want to cover it with logs?

docker-compose.dev.yml
```diff
@@ -0,0 +1,72 @@
...
+    environment:
+      - PORT=${KEYS_API_PORT}
+      - CORS_WHITELIST_REGEXP=${KEYS_API_CORS_WHITELIST_REGEXP}
+      - GLOBAL_THROTTLE_TTL=${KEYS_API_GLOBAL_THROTTLE_TTL}
+      - GLOBAL_THROTTLE_LIMIT=${KEYS_API_GLOBAL_THROTTLE_LIMIT}
+      - GLOBAL_CACHE_TTL=${KEYS_API_GLOBAL_CACHE_TTL}
+      - LOG_LEVEL=${LOG_LEVEL}
+      - LOG_FORMAT=${LOG_FORMAT}
+      - CHAIN_ID=${KEYS_API_CHAIN_ID}
+      - PROVIDERS_URLS=${RPC_URL}
```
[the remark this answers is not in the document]
Eddort: We walk in circles. I initially separated the application config from the Keys-API, then merged it. Now it is obvious that it would be easier to separate and then refactor this part in the context of a separate task.

docker-compose.dev.yml
```diff
@@ -0,0 +1,72 @@
...
+      - DB_NAME=${KEYS_API_DB_NAME}
+      - DB_PORT=5432
```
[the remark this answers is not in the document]
Eddort: Not really. It is inside the docker network. I did it right.

src/guardian/guardian.service.spec.ts
```diff
@@ -1,31 +1,57 @@
...
+    },
+  ],
+  elBlockSnapshot: {
+    blockNumber: 0,
+    blockHash: 'string',
+    timestamp: 0,
+  },
+};
+
 describe('GuardianService', () => {
```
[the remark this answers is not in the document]
Eddort: TODO

docker-compose.dev.yml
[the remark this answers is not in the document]
Eddort: TODO (after testing)

avsetsin (approved):

PR #79, opened 2023-01-16
feat: dsm updates

Eddort:


src/contracts/registry/registry.service.ts
```diff
@@ -118,6 +118,7 @@ export class RegistryService {
   public async getPubkeyLength(): Promise<number> {
     if (!this.cachedPubKeyLength) {
       const contract = await this.repositoryService.getCachedRegistryContract();
+      // TODO: after keys-api take from local constant
```
avsetsin, line 121: I guess the whole Registry module will not be needed at all after the keys API is implemented

src/contracts/security/security.service.spec.ts
```diff
@@ -223,11 +225,11 @@ describe('SecurityService', () => {
       const mockProviderCall = jest
         .spyOn(providerService.provider, 'call')
         .mockImplementation(async () => {
-          const iface = new Interface(SecurityAbi__factory.abi);
-          return iface.encodeFunctionResult('isPaused', [expected]);
+          const iface = new Interface(StakingRouterAbi__factory.abi);
+          return iface.encodeFunctionResult('isActive', [expected]);
```
avsetsin, line 229: Does this test pass? the request calls `isActive` method instead of `isPaused`, there should be reverse logic here: `iface.encodeFunctionResult('isActive', [!expected]);`

src/contracts/security/security.service.ts
```diff
@@ -161,6 +171,7 @@ export class SecurityService {
   @OneAtTime()
   public async pauseDeposits(
```
avsetsin, line 172: In this service we also need to modify methods:
- signDepositData
- signPauseData

src/guardian/guardian.service.ts
```diff
@@ -102,7 +103,7 @@ export class GuardianService implements OnModuleInit {
    */
   public subscribeToEthereumUpdates() {
     const provider = this.providerService.provider;
-
+    // TODO: keys-api — cron job
```
avsetsin, line 106: We also need a loop for modules

src/contracts/security/security.service.ts
```diff
@@ -161,6 +171,7 @@ export class SecurityService {
...
     blockNumber: number,
+    stakingModuleId: number,
```
avsetsin, line 174: need to add it to jsdoc

src/contracts/registry/registry.service.ts
Eddort: I understand, but I want to delete this in keys-API branch.

PR #85, opened 2023-02-20
Test: Ganache e2e main attack scenario test

kolyasapphire:


test/manifest.e2e-spec.ts
```diff
@@ -0,0 +1,382 @@
...
+  });
+
+  afterAll(async () => {
+    await server.close();
+  });
+
+  beforeAll(async () => {
+    // Prepare a signer for the unlocked Ganache account
+    if (!process.env.WALLET_PRIVATE_KEY) {
+      process.exit();
```
avsetsin, line 220: It may not be obvious what the problem is. Can we throw an exception with details?

test/manifest.e2e-spec.ts
```diff
@@ -0,0 +1,382 @@
...
+      await new Promise((res) => setTimeout(res, 1000));
+
+      // Check if on pause now
+      const routerContract = StakingRouterAbi__factory.connect(
+        STAKING_ROUTER,
+        providerService.provider,
+      );
+      const isOnPause = await routerContract.getStakingModuleIsDepositsPaused(
+        1,
+      );
```
avsetsin, line 376: We should check that the module has not been paused before the deposit

avsetsin (approved):
lgtm

test/manifest.e2e-spec.ts
kolyasapphire: Oh yeah, good point! Adding a check

test/manifest.e2e-spec.ts
kolyasapphire: Agreed, adding

PR #99, opened 2023-03-23
Added loading sensitive variables from a file

Kron0S:
Adds a variables `RABBITMQ_PASSCODE_FILE` , `WALLET_PRIVATE_KEY_FILE` , which point to the file from which to read private value for this variable


selfuryon:
We use Kubernetes and providing sensitive information via environment variables is not safe, we need to have a different way to use sensitive vars: we decided to use files as the source, so this PR adds supports for that.

Eddort:
Thank you for your contribution.
I will add it as soon as possible

eddort (approved):

PR #139, opened 2023-07-22
Split cache into files

avsetsin:


src/cache/cache.service.ts
```diff
@@ -44,34 +50,89 @@ export class CacheService<T extends unknown> {
...
+  private async getCacheFilePaths(): Promise<string[]> {
     const dir = await this.getCacheDirPath();
-    return join(dir, this.cacheFile);
+    const existCacheFiles: string[] = [];
+
+    let batchIndex = 0;
+
+    do {
+      try {
+        const filePath = join(dir, this.getCacheFileName(batchIndex));
```
eddort, line 65: Perhaps we should use glob to find the files we need and exclude others?

src/cache/cache.service.ts
```diff
@@ -44,34 +50,89 @@ export class CacheService<T extends unknown> {
...
+      const to = (batchIndex + 1) * this.cacheBatchSize;
+      const batchedData = data.slice(from, to);
+
+      if (batchedData.length === 0) return;
+
+      const filePath = join(dirPath, this.getCacheFileName(batchIndex));
+      await writeFile(filePath, JSON.stringify({ headers, data: batchedData }));
+
+      batchIndex++;
+    } while (true);
```
eddort, line 129: I would suggest using a stricter loop, without the "while true".
And we need to consider the data integrity scenario. I suggest specifying the number of chunks in the headers.

infloop:
why we need force push here?

src/cache/cache.service.ts
```diff
@@ -1,40 +1,46 @@
...
+      this.cache = await this.getCacheFromFiles();
     }
 
     return this.cache;
   }
 
   public async setCache(cache: T): Promise<void> {
     this.cache = cache;
-    return await this.saveCacheToFile();
+    return await this.saveCacheToFiles();
```
infloop, line 38: Maybe add removal of cache files before actual saving of new files?

src/cache/cache.service.ts
avsetsin: replaced with a for loop

src/cache/cache.service.ts
avsetsin: done in `saveCacheToFiles` method

src/cache/cache.service.ts
avsetsin: replaced with `glob`

infloop (approved):

PR #144, opened 2023-08-08
Cache improvements

infloop:


src/cache/cache.service.ts
```diff
@@ -1,128 +1,234 @@
...
+        throw new CacheError('Cache headers are not equal');
+      }
 
-    const dirPath = await this.getCacheDirPath();
-    await mkdir(dirPath, { recursive: true });
+      // checking that all stats equal
+      const allStatsAreEqual = allContents.every(
+        (content, i, arr) =>
+          JSON.stringify(arr[0]?.stats?.dataTotalLength ?? 0) ===
+          JSON.stringify(content.stats.dataTotalLength),
```
eddort, line 185: Is JSON.stringify definitely needed here?

package.json
```diff
@@ -85,7 +86,7 @@
     "ts-node": "^10.0.0",
     "tsconfig-paths": "^3.10.1",
     "typechain": "^5.1.2",
-    "typescript": "^4.3.5"
+    "typescript": "^4.9.5"
```
avsetsin, line 89: looks like the new ts vertsion broke linting

src/cache/cache.service.ts
```diff
@@ -1,128 +1,234 @@
...
+    const res = valueType.safeParse(parsed);
+
+    if (res.success) {
+      return res.data;
+    }
+
+    throw new CacheError(
+      `Cache file [${basename(filePath)}] integrity error`,
+      res.error,
+    );
```
avsetsin, line 232: i would suggest to follow one code style and throw the error in the condition instead of the early exit

avsetsin: I would suggest putting this parsing in a separate method and not passing `valueType` through multiple methods via arguments, but getting it from `this.cacheValueType` in place

package.json
infloop: Fixed

src/cache/cache.service.ts
```diff
@@ -1,128 +1,234 @@
...
+      );
 
-    await this.deleteCacheFiles();
+      if (!allStatsAreEqual) {
+        throw new CacheError('Cache stats are not equal');
+      }
 
-    const totalBatches = Math.ceil(data.length / this.cacheBatchSize);
+      const headers: Headers = allContents[0].headers;
+      const stats = allContents[0].stats;
```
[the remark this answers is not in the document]
infloop: Tried this, it's very hard in that case to separate stats and headers from other part of application.

src/cache/cache.service.ts
infloop: I agree on throwing error on condition. Also I think that passing `valueType` thought multiple methods is important here, because `valueType` argument here is not the same as `this.cacheValueType` that is used in class variable. Argument `valueType` includes `stats`, and `this.cacheValueType` not.

infloop: UPD: re-checked again, the second sentence is not true, `this.cacheValueType` and `valueType` argument are the same

eddort:
@infloop we need to finish this task

src/cache/utils/delete-all-files-from-cache-dir.ts
```diff
@@ -0,0 +1,21 @@
...
+  const absoluteFilePaths = await glob(`*`, {
+    cwd: cacheDir,
+    absolute: true,
+  });
+
+  try {
+    await Promise.all(
+      absoluteFilePaths.map(async (filePath) => unlink(filePath)),
+    );
+  } catch (error) {}
```
eddort, line 20: What if we catch a lack of permissions error in this case? Maybe we should output several types of errors to the console?
```js
switch (errorCode) {
  case 'EACCES':
    throw new Error(
      `Permission denied when trying to read the file specified by ${envVarFile}`,
    );
  case 'EMFILE':
    throw new Error(
      `Too many open files in the system when trying to read the file specified by ${envVarFile}`,
    );
  default:
    return
}
```

eddort:
I'm worried about performance (since we use Zod). Have you done a benchmark?

PR #150, opened 2023-10-15
test improvement

eddort:


cbates8979:
States smart contract out of energy why is it in smart contract form can't it be simple wallet ?

gen-kapi-types.sh
infloop, line 1: Is shebang `#!/bin/sh` needed here?

src/keys-api/interfaces/index.ts
infloop, line 1: After PR https://github.com/lidofinance/lido-keys-api/pull/266 that fixes Swagger types there is a chance that NonNullable types are not needed.

gen-kapi-types.sh
tadhaQ-nerfed-5misbu: What is bin? File Perhaps?

src/keys-api/interfaces/index.ts
tadhaQ-nerfed-5misbu: I don't understand I do have a file like key thingy that opens wallet from file and I have all my keys

PR #152, opened 2023-11-28
feat: duplicated vetted keys checks

Amuhar:


src/guardian/guardian.service.ts
```diff
@@ -136,14 +137,14 @@ export class GuardianService implements OnModuleInit {
...
 
-      await Promise.all(
-        stakingModules.map(async (stakingRouterModule) => {
-          const stakingModuleData =
-            await this.stakingModuleGuardService.getStakingRouterModuleData(
-              stakingRouterModule,
-              blockHash,
-            );
+      // TODO: check only if one of nonce changed
+      await this.stakingModuleGuardService.checkVettedKeysDuplicates(
```
eddort, line 141: in this code, if there is a key crossing, it will stop all deposits, is this the expected behavior?

eddort: perhaps we should reconsider here and skip deposits only in modules where there are overlaps

src/guardian/staking-module-guard/staking-module-guard.service.ts
```diff
@@ -30,34 +31,55 @@ export class StakingModuleGuardService {
...
 
-    return {
-      nonce,
-      unusedKeys: keys.map((srKey) => srKey.key),
-      isDepositsPaused,
-      stakingModuleId: stakingRouterModule.id,
-      blockHash,
-    };
+    if (duplicatedKeys.length) {
+      this.logger.warn('Found duplicated vetted key', {
```
eddort, line 76: I would suggest additionally log the module's ID as well

src/staking-router/staking-router.service.ts
```diff
@@ -13,6 +15,77 @@ export class StakingRouterService {
...
+    const operatorsByModules =
+      await this.keysApiService.getOperatorListWithModule();
+    const operatorsBlockHash =
+      operatorsByModules.meta.elBlockSnapshot.blockHash;
+    const operatorsBlockNumber =
+      operatorsByModules.meta.elBlockSnapshot.blockNumber;
+
+    const unusedKeys = await this.keysApiService.getUnusedKeys();
+    const keysBlockHash = unusedKeys.meta.elBlockSnapshot.blockHash;
+    if (keysBlockHash != operatorsBlockHash) {
```
eddort, line 29: it's always a good idea to use a strict comparison `!==`

src/staking-router/staking-router.service.ts
```diff
@@ -13,6 +15,77 @@ export class StakingRouterService {
...
+  public async getVettedAndUnusedKeys() {
+    // TODO: add cache by modules nonce
+    const operatorsByModules =
+      await this.keysApiService.getOperatorListWithModule();
+    const operatorsBlockHash =
+      operatorsByModules.meta.elBlockSnapshot.blockHash;
+    const operatorsBlockNumber =
+      operatorsByModules.meta.elBlockSnapshot.blockNumber;
+
+    const unusedKeys = await this.keysApiService.getUnusedKeys();
```
eddort, line 27: If we decided to load all data at once, we need to calculate at what point in time memory will be insufficient and what to do in this situation

src/staking-router/staking-router.service.ts
```diff
@@ -13,6 +15,77 @@ export class StakingRouterService {
     protected readonly keysApiService: KeysApiService,
   ) {}
 
+  public async getVettedAndUnusedKeys() {
```
eddort, line 18: I suggest making the endpoint on the Keys-API side. it can be done in SQL using sorting and $lt $gt operators

src/guardian/staking-module-guard/staking-module-guard.service.ts
```diff
@@ -156,6 +196,29 @@ export class StakingModuleGuardService {
     return attackIntersections;
   }
 
+  public async handleIntersectionBetweenUsedAndUnusedKeys(
```
eddort, line 199: I would suggest to name the method something else, since it does not process this case in any way, but returns duplicates and outputs the log

src/contracts/deposit/deposit.service.ts
```diff
@@ -36,6 +36,7 @@ export class DepositService {
     private blsService: BlsService,
   ) {}
 
+  // Why OneAtTime() ? this function is executed in guardian.service inside another function with OneAtTime() in cron job
```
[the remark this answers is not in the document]
Amuhar: ok, deleted

src/guardian/guardian.service.ts
Amuhar: you are write, i wrote it before decided to skip deposits only for one module
will fix

src/staking-router/staking-router.service.ts
Amuhar: blockHash of two requests is checking down as it was before in getStakingModuleUnusedKeys method

Maybe we could consider here use of /v2/modules/operators/keys endpoint, but than we need to add 'used' filter
because at the moment we could fetch only all keys

src/staking-router/staking-router.service.ts
Amuhar: i am okay with it but i am just not sure we have enough time to make fixes in two applications

src/staking-router/staking-router.service.ts
Amuhar: added task for it https://linear.app/lidofi/issue/VAL-493/solving-problem-with-vetted-keys-fetching

src/staking-router/staking-router.service.ts
Amuhar: it is related to  https://linear.app/lidofi/issue/VAL-493/solving-problem-with-vetted-keys-fetching

src/staking-router/staking-router.service.ts
Amuhar: fixed, thank you

src/guardian/staking-module-guard/staking-module-guard.service.ts
Amuhar: agree

src/guardian/staking-module-guard/staking-module-guard.service.ts
```diff
@@ -156,6 +227,43 @@ export class StakingModuleGuardService {
...
 
+  public async getIntersectionBetweenUsedAndUnusedKeys(
+    intersectionsWithLidoWC: VerifiedDepositEvent[],
+    blockData: BlockData,
+  ) {
+    const depositedPubkeys = intersectionsWithLidoWC.map(
+      (deposit) => deposit.pubkey,
+    );
+
+    if (depositedPubkeys.length) {
```
eddort, line 238: I would suggest inverting the condition and making an early exit

eddort:
It looks good, but I want to take another look tomorrow morning

src/guardian/staking-module-guard/staking-module-guard.service.ts
```diff
@@ -30,34 +29,85 @@ export class StakingModuleGuardService {
...
+      });
 
-    return {
-      nonce,
-      unusedKeys: keys.map((srKey) => srKey.key),
-      isDepositsPaused,
-      stakingModuleId: stakingRouterModule.id,
-      blockHash,
-    };
+      //TODO: set prometheus metric council_daemon_vetted_unused_duplicate
```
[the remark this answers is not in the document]
Amuhar: we have a task for it

src/staking-router/staking-router.service.ts
Amuhar: @eddort please, resolve conversations :)

src/guardian/staking-module-guard/staking-module-guard.service.ts
Amuhar: okay

src/guardian/staking-module-guard/staking-module-guard.service.ts
```diff
@@ -156,6 +227,43 @@ export class StakingModuleGuardService {
...
+    }
+
+    this.logger.log(
+      'Found intersections with lido credentials, need to check duplicated keys',
+    );
+
+    const { data, meta } =
+      await this.stakingRouterService.getKeysWithDuplicates(depositedPubkeys);
+
+    if (meta.elBlockSnapshot.blockNumber < blockData.blockNumber) {
```
eddort, line 249: It would be good to add a `blockHash` check as was done earlier

```js
  public isNeedToProcessNewState(newMeta: {
    blockHash: string;
    blockNumber: number;
  }) {
    const lastMeta = this.lastProcessedStateMeta;
    if (!lastMeta) return true;
    if (lastMeta.blockNumber > newMeta.blockNumber) {
      this.logger.error('Keys-api returns old state', newMeta);
      return false;
    }
    return lastMeta.blockHash !== newMeta.blockHash;
  }
```

src/guardian/staking-module-guard/staking-module-guard.service.ts
Amuhar: in log will be a list of keys with modules, there they were found

src/guardian/staking-module-guard/staking-module-guard.service.ts
Amuhar: i think code will be easier to read if we use oldMeta from method parameter
i dont see here a reason for cache this value

Amuhar: but for readability i will add function for condition

Amuhar: as we know that used keys shoudnt be deleted

eddort: Is the blockHash check not needed here?

src/guardian/staking-module-guard/staking-module-guard.service.ts
Amuhar: @eddort unresolved conv

src/keys-api/interfaces/RegistryOperator.ts
```diff
@@ -0,0 +1,38 @@
...
+  /**
+   * This value shows if node operator active
+   */
+  active: boolean;
+  /**
+   * Operator name
+   */
+  name: string;
+  /**
+   * Ethereum 1 address which receives stETH rewards for this operator
```
F4ever, line 15:
```suggestion
   * Ethereum address which receives stETH rewards for this operator
```

src/staking-router/staking-router.service.ts
```diff
@@ -13,6 +16,141 @@ export class StakingRouterService {
...
+    if (keysBlockHash !== operatorsBlockHash) {
+      this.logger.error(
+        'Blockhash of the received keys and operators dont match',
+        {
+          keysBlockHash,
+          operatorsBlockHash,
+        },
+      );
+
+      throw new Error(
```
F4ever, line 84: Specify Error.
smth like `throw new InconsistentDataError(`
And in `getOperatorsAndKeysFromKAPI` method if you catch this error try to refetch first request

F4ever: You can do it few times until get Consistent data or maxRetryExceed

src/guardian/guardian.service.ts
```diff
@@ -136,14 +137,25 @@ export class GuardianService implements OnModuleInit {
...
 
-      await Promise.all(
-        stakingModules.map(async (stakingRouterModule) => {
-          const stakingModuleData =
-            await this.stakingModuleGuardService.getStakingRouterModuleData(
-              stakingRouterModule,
-              blockHash,
-            );
+      // TODO: check only if one of nonce changed
+      const modulesIdWithDuplicateKeys: number[] =
```
F4ever, line 141: `handleNewBlock` function is too big.
This function is responsible only to check if are there any updates or not.
Let's move logic below to new function `proceedModules(blockData)`

F4ever: And ping massage generator to `sendPingToBroker(blockData)`

src/guardian/guardian.service.ts
```diff
@@ -96,10 +97,8 @@ export class GuardianService implements OnModuleInit {
     this.logger.log('New staking router state cycle start');
 
     try {
-      const {
-        elBlockSnapshot: { blockHash, blockNumber },
-        data: stakingModules,
-      } = await this.stakingRouterService.getStakingModules();
+      const { blockHash, blockNumber, stakingModulesData } =
```
F4ever, line 100: let's separate business logic from non-business.
Move all try content into separate function

src/guardian/guardian.service.ts
```diff
@@ -96,10 +97,8 @@ export class GuardianService implements OnModuleInit {
...
+        await this.stakingRouterService.getVettedAndUnusedKeys();
```
F4ever, line 101: The result of this function is not a VettedAndUnusedKeys, but stakingModulesData

src/guardian/guardian.service.ts
F4ever: This function should only check if new block shuld be proceed or not

F4ever: I see it  like this
`handleNewBlock(){
    meta = this.keysApiService.getLatestMeta();
    if (this.blockGuardService.isNeedToProcessNewState(meta))
    { try {
         createAndSendDepositSignature()
    } except {}
}`

src/guardian/guardian.service.ts
```diff
@@ -119,8 +118,10 @@ export class GuardianService implements OnModuleInit {
         return;
       }
 
+      const stakingModulesNumber = stakingModulesData.length;
```
F4ever, line 121: stakingModuleLength is better

src/guardian/guardian.service.ts
```diff
@@ -136,14 +137,25 @@ export class GuardianService implements OnModuleInit {
...
-              blockHash,
-            );
+      // TODO: check only if one of nonce changed
+      const modulesIdWithDuplicateKeys: number[] =
+        this.stakingModuleGuardService.checkVettedKeysDuplicates(
+          stakingModulesData,
+          blockData,
+        );
+
+      const stakingModulesWithoutDuplicates: StakingModuleData[] =
```
F4ever, line 147: I would advice to rewrite this function to
-> Fetch all modules and vetted unused keys for it
-> for module in modules
    -> Promiss all
        -> Check duplicates
        -> Check keys validity
        -> Send deposit message

F4ever: Currently function with name `checkKeysIntersections` somewhy send deposit message, but should only check intersections.

src/guardian/guardian.service.ts
```diff
@@ -136,14 +137,25 @@ export class GuardianService implements OnModuleInit {
...
+        this.stakingModuleGuardService.excludeModulesWithDuplicatedKeys(
```
F4ever, line 148: After we will do check duplication for each module separatly this would be reduntant

src/keys-api/keys-api.service.ts
```diff
@@ -54,4 +60,33 @@ export class KeysApiService {
       throw Error('Keys API not synced, please wait');
     return result;
   }
+
+  /**
+   *
+   * @param The /v1/keys/find KAPI endpoint returns a key along with its duplicates
+   * @returns
+   */
+  public async getKeysWithDuplicates(pubkeys: string[]) {
```
F4ever, line 69: It doesn't return duplicates. It returns all keys entities with provided public keys

src/guardian/staking-module-guard/staking-module-guard.service.ts
```diff
@@ -30,34 +30,85 @@ export class StakingModuleGuardService {
...
-        module: { nonce },
-      },
-    } = await this.stakingRouterService.getStakingModuleUnusedKeys(
-      blockHash,
-      stakingRouterModule,
-    );
+  /**
+   * @returns List of staking modules id with duplicates
+   */
+  public checkVettedKeysDuplicates(
```
F4ever, line 36: `getModulesWithDuplicatedKeys`

F4ever: I think we should exclude here modules with duplicated keys where keys are already used

F4ever: -> Create a hash map with all vetted keys
-> for each module
    -> check unused vetted keys is in this hash module
    -> if there is that key, then `duplicatedModulesIds.add(moduleId)`

src/guardian/staking-module-guard/staking-module-guard.service.ts
```diff
@@ -30,34 +30,85 @@ export class StakingModuleGuardService {
...
+    return [];
+  }
+
+  private hasDuplicateKeys(stakingModules: Map<number, number>): boolean {
+    const moduleCounts = Array.from(stakingModules.values());
+
+    return stakingModules.size > 1 || moduleCounts[0] > 1;
+  }
+
+  public excludeModulesWithDuplicatedKeys(
```
F4ever, line 94: Looks like a simple filter that removes modules by id
`modules.filter((module)=>duplicatedModules.indexOf(stakingModuleId) != -1)`
Why so much code)

src/guardian/guardian.service.ts
Amuhar: fixed

src/keys-api/keys-api.service.ts
Amuhar: it return keys with duplicates

src/guardian/staking-module-guard/staking-module-guard.service.ts
Amuhar: we discussed it already and agreed not to check used vetted keys 
we have this information in deposited events, and check only if we have need to do it

src/guardian/guardian.service.ts
Amuhar: i dont think your algorithm is right
we check duplicated keys across all modules and exclude modules with duplicates

Amuhar: and you skip intersection check as well
Promis.all also does make sense to put inside for module in modules

src/guardian/guardian.service.ts
Amuhar: this i can fix

src/staking-router/staking-router.service.ts
Amuhar: add class to error i can 
retry i will try to consider, thank you

src/guardian/staking-module-guard/staking-module-guard.service.ts
Amuhar: it is enough  return stakingModulesData.filter(
        ({ stakingModuleId }) =>
          !modulesIdWithDuplicateKeys.includes(stakingModuleId),
      );

src/guardian/guardian.service.ts
Amuhar: i dont think check separatly duplicates is a good idea
we look for duplicates across modules

src/guardian/guardian.service.ts
Amuhar: checkKeysIntersections did it before, this pr didnt change it
   refactoring of council daemon could be done anytime later

src/guardian/staking-module-guard/staking-module-guard.service.ts
F4ever: VettedKeys -> UnusedVettedKeys

src/keys-api/keys-api.service.ts
F4ever: Maybe findKeyEntities or Entries

src/guardian/staking-module-guard/staking-module-guard.service.ts
F4ever: All resolved except function name and vetted keys are unsued vetted

src/keys-api/keys-api.service.ts
Amuhar: fixed

src/guardian/staking-module-guard/staking-module-guard.service.ts
Amuhar: fixed

src/staking-router/staking-router.service.ts
```diff
@@ -2,63 +2,104 @@ import { Injectable, LoggerService, Inject } from '@nestjs/common';
...
 
-  protected getStakingRouterKeysCache(stakingModuleId: number) {
-    return this.stakingRouterCache[stakingModuleId];
+    return { stakingModulesData, blockHash, blockNumber };
   }
 
-  protected setStakingRouterCache(
-    stakingModuleId: number,
-    srResponse: SRModuleKeysResponse,
+  private isEqualLastChangedBlockHash(
```
eddort, line 58: are the two methods the same?

src/guardian/staking-module-guard/staking-module-guard.service.ts
```diff
@@ -131,29 +201,71 @@ export class StakingModuleGuardService {
...
+
+    this.isEqualLastChangedBlockHash(
+      lastChangedBlockHash,
+      meta.elBlockSnapshot.lastChangedBlockHash,
+    );
+
+    return data.filter((key) => key.used);
+  }
+
+  private isEqualLastChangedBlockHash(
```
eddort, line 257: are the two methods the same?

src/staking-router/vetted-keys.ts
```diff
@@ -0,0 +1,16 @@
+import { RegistryKey } from 'keys-api/interfaces/RegistryKey';
+import { RegistryOperator } from 'keys-api/interfaces/RegistryOperator';
+
+export function getVettedUnusedKeys(
```
eddort, line 4: perhaps we should write a small test for this function?

src/staking-router/staking-router.service.ts
```diff
@@ -2,63 +2,104 @@ import { Injectable, LoggerService, Inject } from '@nestjs/common';
...
+import { RegistryOperator } from 'keys-api/interfaces/RegistryOperator';
+import { RegistryKey } from 'keys-api/interfaces/RegistryKey';
+import { SRModule } from 'keys-api/interfaces';
+import { Meta } from 'keys-api/interfaces/Meta';
+import { InconsistentLastChangedBlockHash } from 'common/custom-errors';
 
 @Injectable()
 export class StakingRouterService {
-  protected stakingRouterCache: Record<number, SRModuleKeysResponse> = {};
+  // protected stakingRouterCache: Record<number, StakingModuleData> = {};
```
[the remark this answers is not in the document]
Amuhar: i will make this solution today

src/staking-router/staking-router.service.ts
Amuhar: you are right, will reuse

src/staking-router/vetted-keys.ts
Amuhar: yes, will do

src/guardian/guardian-metrics/guardian-metrics.service.ts
```diff
@@ -1,4 +1,4 @@
-import { Injectable } from '@nestjs/common';
+import { Inject, Injectable } from '@nestjs/common';
```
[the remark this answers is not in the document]
Amuhar: will fix

src/staking-router/staking-router.service.ts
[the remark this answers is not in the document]
Amuhar: will prepare it in a separate branch

eddort (approved):

eddort:
It would be nice to write more tests for key validation in the future

eddort (approved):

src/guardian/staking-module-guard/staking-module-guard.service.ts
```diff
@@ -433,7 +436,15 @@ export class StakingModuleGuardService {
   ): boolean {
     if (!firstState || !secondState) return false;
     if (firstState.depositRoot !== secondState.depositRoot) return false;
+
+    // Check if the nonce values are different. A difference in nonce implies a state change.
     if (firstState.nonce !== secondState.nonce) return false;
```
F4ever: Do we still need this check? Seems block_hash check is enough.

Amuhar: you are right , we dont need it. but file you check is outdated

PR #155, opened 2023-12-12
e2e for duplicated keys check and invalid keys

Amuhar:


src/contracts/deposit/deposit.service.ts
```diff
@@ -243,6 +243,7 @@ export class DepositService {
    * The last N blocks are not stored, in order to avoid storing reorganized blocks
    */
   public async updateEventsCache(): Promise<void> {
+    this.logger.log('try to update?');
```
F4ever, line 246: Fix logs

src/contracts/deposit/deposit.service.ts
```diff
@@ -270,6 +271,8 @@ export class DepositService {
         chunkToBlock,
       );
 
+      console.log('chunk?', chunkEventGroup);
```
F4ever, line 274: Fix logs

F4ever:
logs fixed in next prs 😬

F4ever (approved):

PR #158, opened 2023-12-14
fix: added cache

Amuhar:


src/guardian/keys-validation/keys-validation.service.ts
```diff
@@ -27,29 +43,54 @@ export class KeysValidationService {
...
+
+  prepareKeyForValidation(
+    key: RegistryKey,
+    withdrawalCredentials: string,
+    forkVersion: Uint8Array,
+  ) {
+    const cachedEntry = this.keysCache.get(key.key);
+
+    // key wasn't in cache or signature was changed
+    if (cachedEntry && cachedEntry.signature == key.depositSignature) {
```
F4ever, line 76: prepare function can return null. Rename this function to filter and use it like filter

src/guardian/keys-validation/keys-validation.service.ts
```diff
@@ -27,29 +43,63 @@ export class KeysValidationService {
...
+  filterKeyForValidation(
+    key: RegistryKey,
+    withdrawalCredentials: string,
+    forkVersion: Uint8Array,
+  ): DepositData | null {
+    const cachedEntry = this.keysCache.get(key.key);
+
+    // key was in cache and signature wasn't changed
+    if (this.keyNeedToBeValidated(cachedEntry, key)) {
+      return null;
```
F4ever, line 79: return false

src/guardian/keys-validation/keys-validation.service.ts
```diff
@@ -27,29 +43,63 @@ export class KeysValidationService {
...
+    }
+
+    return this.depositData(key, withdrawalCredentials, forkVersion);
```
F4ever, line 82: return true

src/guardian/keys-validation/keys-validation.service.ts
```diff
@@ -27,29 +43,63 @@ export class KeysValidationService {
...
-      vettedKeys.map((key) => ({
-        key: key.key,
-        depositSignature: key.depositSignature,
-        used: false,
-        withdrawalCredentials: bufferFromHexString(withdrawalCredentials),
-        genesisForkVersion: Buffer.from(forkVersion.buffer),
-      })),
-    );
+    const keysForValidation = vettedKeys
+      .map((key) =>
```
F4ever, line 47: `.filter(this.filterKeyForValidation).map((key)=>this.depositData(key, withdrawalCredentials, forkVersion))`

F4ever: filterKeyForValidation - removes all validated keys
depositData - converts all keys to format that is ready for validation

src/guardian/keys-validation/keys-validation.service.ts
```diff
@@ -27,29 +43,63 @@ export class KeysValidationService {
...
+
+    // key was in cache and signature wasn't changed
+    if (this.keyNeedToBeValidated(cachedEntry, key)) {
+      return null;
+    }
+
+    return this.depositData(key, withdrawalCredentials, forkVersion);
+  }
+
+  keyNeedToBeValidated(
```
F4ever, line 85: rename to `isKeyValidated`
This function tells if the provided key already was validated

Without `function` word is hard to understand if it is implementation or just function call

src/guardian/keys-validation/keys-validation.service.ts
```diff
@@ -27,29 +43,63 @@ export class KeysValidationService {
...
+    cachedEntry: { signature: string; isValid: boolean } | undefined,
+    key: RegistryKey,
+  ): boolean {
+    return !!cachedEntry && cachedEntry.signature == key.depositSignature;
+  }
+
+  depositData(
```
F4ever, line 92: Maybe `toDepositData`. 
Now it looks like variable, not a function

src/guardian/keys-validation/keys-validation.service.ts
```diff
@@ -62,4 +112,53 @@ export class KeysValidationService {
 
     return forkVersion;
   }
+
+  async updateCacheWithValidationResults(
```
F4ever, line 116: maybe rename to `updateCache`

src/guardian/keys-validation/keys-validation.service.ts
Amuhar: It is a method .

src/guardian/keys-validation/keys-validation.service.ts
```diff
@@ -27,29 +43,63 @@ export class KeysValidationService {
...
+    }
+
+    return this.depositData(key, withdrawalCredentials, forkVersion);
+  }
+
+  keyNeedToBeValidated(
+    cachedEntry: { signature: string; isValid: boolean } | undefined,
+    key: RegistryKey,
+  ): boolean {
+    return !!cachedEntry && cachedEntry.signature == key.depositSignature;
```
F4ever, line 89: You have to check all key object, not only signature.

Bad scenario
User uploads key.
We validate that key
He deletes that key and uploads with same signature and pubkey, but with wrong creds or other field.
Signature is broken, but we think it's ok, because we already validated it

src/guardian/keys-validation/keys-validation.service.ts
Amuhar: It a method

src/guardian/keys-validation/keys-validation.service.ts
```diff
@@ -27,29 +43,63 @@ export class KeysValidationService {
   ): Promise<{ key: string; depositSignature: string }[]> {
```
F4ever, line 43: Call this function `findInvalidKeys`

F4ever: Function returns only invalid keys

src/guardian/keys-validation/keys-validation.service.ts
```diff
@@ -62,4 +112,53 @@ export class KeysValidationService {
...
+  ) {
+    validatedKeys.forEach(([key, isValid]) =>
+      this.keysCache.set(key.key, { signature: key.depositSignature, isValid }),
+    );
+  }
+
+  private mergeInvalidKeys(
+    vettedKeys: RegistryKey[],
+    recentInvalidKeys: { key: string; depositSignature: string }[],
+  ): { key: string; depositSignature: string }[] {
```
F4ever, line 127: 1. Excude checked keys (in cache)
2. Validate keys
3. Write to cache
4. Cachec is overfilled. Lastest keys are deleted from cache.
5.  There were some invalid keys
6. Council daemon send message even it shouldn't

F4ever: I propose to rewrite and do next things
1. Get results from cache filtered by vetted keys
2. Filter not validated keys and validated them
3. merge two lists

src/guardian/keys-validation/keys-validation.service.ts
```diff
@@ -27,29 +43,54 @@ export class KeysValidationService {
...
       }));
+
+    return this.mergeInvalidKeys(vettedKeys, invalidKeysFromCurrentValidation);
+  }
+
+  prepareKeyForValidation(
+    key: RegistryKey,
+    withdrawalCredentials: string,
+    forkVersion: Uint8Array,
+  ) {
```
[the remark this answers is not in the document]
Amuhar: fixed

src/guardian/keys-validation/keys-validation.service.ts
Amuhar: fixed

src/guardian/keys-validation/keys-validation.service.ts
Amuhar: fixed

src/guardian/keys-validation/keys-validation.service.ts
Amuhar: fixed

src/guardian/keys-validation/keys-validation.service.ts
Amuhar: fixed

src/guardian/keys-validation/keys-validation.service.ts
Amuhar: renamed method name

src/guardian/keys-validation/keys-validation.service.ts
Amuhar: renamed

src/guardian/keys-validation/keys-validation.service.ts
Amuhar: fixed

src/guardian/keys-validation/keys-validation.service.ts
Amuhar: agree, here could be a problem

F4ever (approved):

PR #161, opened 2023-12-26
 kapi resp cache

Amuhar:


src/guardian/staking-module-guard/staking-module-guard.service.ts
```diff
@@ -28,6 +28,7 @@ export class StakingModuleGuardService {
 
   private lastContractsStateByModuleId: Record<number, ContractsState | null> =
     {};
+  private duplicatedKeysWasFound = false;
```
eddort, line 31: it's not done yet?

Amuhar: Yes, about this optimization I was speaking
there are no problem to release now without it

Amuhar: check for finding duplicated keys is already implemented 
there are possibility to not make it on each iteration

Amuhar: check for finding duplicated keys is already implemented - and released on holesky

src/staking-router/staking-router.service.ts
```diff
@@ -18,39 +19,88 @@ export class StakingRouterService {
...
+
+    // if module was deleted we will not include data in final result
+    // TODO: add test on delete
+    operatorsByModules.forEach(({ module: stakingModuleData }) => {
+      if (this.isDataOutdated(stakingModuleData)) {
+        outdatedModuleAddresses.push(stakingModuleData.stakingModuleAddress);
+        return;
+      }
+      cachedData.push(this.stakingRouterCache[stakingModuleData.id]);
+      return;
```
eddort, line 57: unnecessary return

src/staking-router/staking-router.service.ts
```diff
@@ -18,39 +19,88 @@ export class StakingRouterService {
...
           operators,
           stakingModule,
           unusedKeys,
           meta: operatorsMeta,
         }),
-    );
+      );
+
+    // update cache
+    actualizedData.forEach((stakingModule: StakingModuleData) => {
```
eddort, line 86: there could be a cache overflow

src/keys-api/keys-api.service.ts
```diff
@@ -57,8 +58,14 @@ export class KeysApiService {
     return result;
   }
 
-  public async getUnusedKeys() {
-    const result = await this.fetch<KeyListResponse>(`/v1/keys?used=false`);
+  public async getUnusedKeys(moduleAddresses?: string[]) {
```
eddort, line 61: why moduleAddresses can be undefined?

src/guardian/staking-module-guard/staking-module-guard.service.ts
```diff
@@ -300,6 +313,98 @@ export class StakingModuleGuardService {
...
+  public async getInvalidKeys(
+    stakingModuleData: StakingModuleData,
+    blockData: BlockData,
+  ): Promise<{ key: string; depositSignature: string }[]> {
+    this.logger.log('Start keys validation', {
+      keysCount: stakingModuleData.vettedUnusedKeys.length,
+      moduleId: stakingModuleData.stakingModuleId,
+    });
+    const validationTimeStart = performance.now();
+    const invalidKeysList = await this.keysValidationService.findInvalidKeys(
```
eddort, line 391: it is not very convenient to read the findInvalidKeys and getInvalidKeys methods. While reading it seems that this is a request to the server, but not validation. Perhaps they should be renamed?

src/staking-router/staking-router.service.ts
```diff
@@ -18,39 +19,88 @@ export class StakingRouterService {
...
 
-    const blockHash = operatorsMeta.elBlockSnapshot.blockHash;
-    const blockNumber = operatorsMeta.elBlockSnapshot.blockNumber;
+    return { data: unusedKeys, meta: unusedKeysMeta };
+  }
+
+  /**
+   * Return staking module data and block information
+   */
+  async getStakingModulesData(
```
eddort, line 41: This is essentially a caching layer around the method of retrieving unused keys. perhaps we should reflect that in the name?

src/staking-router/staking-router.service.ts
```diff
@@ -18,39 +19,88 @@ export class StakingRouterService {
...
+      await this.fetchUnusedKeys(outdatedModuleAddresses);
 
     this.isEqualLastChangedBlockHash(
       operatorsMeta.elBlockSnapshot.lastChangedBlockHash,
       unusedKeysMeta.elBlockSnapshot.lastChangedBlockHash,
     );
 
-    const stakingModulesData = operatorsByModules.map(
-      ({ operators, module: stakingModule }) =>
+    const actualizedData = operatorsByModules
```
eddort, line 70: I would suggest greatly simplifying this code, leaving only the keys to be cached

eddort:
After re-reading the code, I found an error: the findAlreadyDepositedKeys method uses the wrong lastChangedBlockHash. It is necessary to use lastChangedBlockHash from EL meta, not from the module

Amuhar:
The mistake in the 'findAlreadyDepositedKeys' method occurred in this PR, not previously. This is because in the 'staking-router-service', the 'processStakingData' method takes and stores the 'lastChangedBlockHash' for 'StakingModuleData', and everything worked as it should. However, in this PR, there might be a problem since part of the staking module data was taken from the cache, potentially leading to an outdated 'lastChangedBlockHash'. The mistake here is caching the 'lastChangedBlockHash' when we really need to cache only the keys.

Regardless, I prefer returning to the plan we had on Monday and not rushing with caching for KAPI. I would prepare a PR with fewer changes.

src/guardian/staking-module-guard/staking-module-guard.service.ts
Amuhar: Good names usually don't come easily. For 'find', it could be 'validate', but there's also a cache, so maybe 'find' is more suitable. As for 'getInvalidKeys', I'm not sure how to rename it right now.

PR #169, opened 2024-01-15
develop -> main

Amuhar:


src/keys-api/interfaces/GroupedByModuleOperatorListResponse.ts
infloop, line 1: Pascal case files are not following naming convention here

src/keys-api/interfaces/KeyListResponse.ts
infloop, line 1: Pascal case file name

src/keys-api/keys-api.service.ts
```diff
@@ -39,19 +41,31 @@ export class KeysApiService {
...
-  public async getModulesList() {
-    const result = await this.fetch<SRModuleListResponse>('/v1/modules');
-    if (!result.data?.length || !result.elBlockSnapshot)
-      throw Error('Keys API not synced, please wait');
+  /**
+   *
+   * @param The /v1/keys/find API endpoint returns keys along with their duplicates
+   * @returns
+   */
+  public async findKeysEntires(pubkeys: string[]) {
```
infloop, line 49: Looks like there is a convention for naming class methods that fetches data from KAPI  - they start with `get` and then follows with `what we are getting`. This method has a typo and starts with `find` and has strange `entries` in the name. I think it's better to rename it to `getKeysByPubkeys`.

src/keys-api/interfaces/GroupedByModuleOperatorListResponse.ts
Amuhar: all interfaces named like this in council, that is why

src/keys-api/interfaces/KeyListResponse.ts
Amuhar: the same

src/keys-api/keys-api.service.ts
Amuhar: okay, will fix

infloop (dismissed):

infloop (approved):

choooze (approved):

PR #197, opened 2024-04-10
feat: deposit cache integrity checks

eddort:


src/contracts/deposit/deposit.service.ts
```diff
@@ -257,21 +238,21 @@ export class DepositService {
    * Updates the cache deposited events
    * The last N blocks are not stored, in order to avoid storing reorganized blocks
    */
-  public async updateEventsCache(): Promise<void> {
+  public async updateEventsCache(): Promise<number> {
     const fetchTimeStart = performance.now();
 
     const [currentBlock, initialCache] = await Promise.all([
+      // TODO: check reorg, add finalized
```
Amuhar, line 245: What is this TODO supposed to do? When will we address it?

src/contracts/deposit/integrity-checker/integrity-checker.service.ts
```diff
@@ -0,0 +1,141 @@
...
+   * @param {VerifiedDepositEvent[]} eventsCache - Events to insert into the tree.
+   */
+  public async putEventsToTree(
+    tree: DepositTree,
+    eventsCache: VerifiedDepositEvent[],
+  ) {
+    for (const [index, event] of eventsCache.entries()) {
+      tree.insertNode(event.depositDataRoot);
+
+      if (index % 200_000 === 0) {
```
Amuhar, line 117: What is 200_000? lets put in a constant

src/contracts/deposit/integrity-checker/integrity-checker.service.ts
```diff
@@ -0,0 +1,141 @@
...
+import { BlockTag } from 'provider';
+import { DepositTree } from '../deposit-tree';
+import {
+  VerifiedDepositEvent,
+  VerifiedDepositEventsCache,
+} from '../interfaces';
+
+@Injectable()
+export class DepositIntegrityCheckerService {
+  finalizedTree = new DepositTree();
```
Amuhar, line 13: Why do we name it finalized? It makes impression that it connects with block finality

src/contracts/deposit/deposit.service.ts
```diff
@@ -11,50 +11,53 @@ import {
...
     if (blockNumber % DEPOSIT_EVENTS_CACHE_UPDATE_BLOCK_RATE !== 0) return;
 
     // The event cache is stored with an N block lag to avoid caching data from uncle blocks
     // so we don't worry about blockHash here
-    await this.updateEventsCache();
+    const toBlockNumber = await this.updateEventsCache();
+    await this.depositIntegrityCheckerService.checkFinalizedRoot(toBlockNumber);
   }
 
   public async initialize(blockNumber: number) {
```
Amuhar, line 47: here we fetch cache twice: for validation of cache age and for update 
is it make sence to send cache header as parameter to updateEventsCache() ?

src/contracts/deposit/interfaces/event.interface.ts
```diff
@@ -7,6 +7,9 @@ export interface DepositEvent {
   blockNumber: number;
   blockHash: string;
   logIndex: number;
+  index: string;
+  depositCount: number;
```
Amuhar, line 11: is it a unique number of event in a deposit contract?

src/contracts/deposit/leveldb/leveldb.service.ts
```diff
@@ -0,0 +1,185 @@
...
+   */
+  private generateDepositKey(number: number): string {
+    if (number < 0 || number > MAX_DEPOSIT_COUNT) {
+      throw new Error(
+        `Deposit count is out of the valid range (0 to ${MAX_DEPOSIT_COUNT}) received ${number}`,
+      );
+    }
+    const index = Buffer.alloc(4);
+    index.writeUInt32BE(number, 0);
+    return `deposit:${index.toString('hex')}`;
```
Amuhar, line 106: Do we need the 'deposit:' prefix? Is it added in case there are different values in the same cache, not just deposit values?

Let's rename generateDepositKey to generateCacheKey. DepositKey sounds like a public key.

src/contracts/deposit/deposit.service.ts
eddort: I had the idea of getting data not once in X blocks, but on each finalized block
I'll come back later and think about it, thanks for pointing it out.

src/contracts/deposit/interfaces/event.interface.ts
eddort: depositCount is an incremental variable that is incremented for each deposit.

src/contracts/deposit/deposit.service.spec.ts
```diff
@@ -357,9 +359,9 @@ describe('DepositService', () => {
         }));
 
       const mockSetCachedEvents = jest
-        .spyOn(depositService, 'setCachedEvents')
+        .spyOn(cacheService, 'insertEventsCacheBatch')
         .mockImplementation(async () => undefined);
-
+      ///!!!
```
Amuhar, line 364: What should we do here?

src/contracts/deposit/deposit.service.spec.ts
```diff
@@ -94,7 +96,7 @@ describe('DepositService', () => {
       expect(blockNumber).toBeGreaterThan(0);
     });
 
-    it('should return block number for mainnet', async () => {
+    it.skip('should return block number for mainnet', async () => {
```
Amuhar, line 99: Why do we skip test?

src/contracts/deposit/deposit.service.ts
eddort: I'm not sure what you mean
Could you explain?

src/contracts/deposit/deposit.service.spec.ts
eddort: this test has stopped passing and I don't understand its meaning. it checks the ethers provider's behavior in a case that has nothing to do with our program.

src/contracts/deposit/deposit.service.ts
```diff
@@ -196,7 +184,8 @@ export class DepositService {
   public async setCachedEvents(
     cachedEvents: VerifiedDepositEventsCache,
   ): Promise<void> {
-    return await this.cacheService.setCache({
+    await this.levelDBCacheService.deleteCache();
+    await this.levelDBCacheService.insertEventsCacheBatch({
       ...cachedEvents,
       headers: {
```
Amuhar, line 190: is setCachedEvents used only in tests?
we still have version here, do we use it somewhere?

src/contracts/deposit/deposit.service.ts
```diff
@@ -285,34 +266,38 @@ export class DepositService {
         chunkToBlock,
       );
 
-      updatedCachedEvents.headers.endBlock = chunkEventGroup.endBlock;
-      updatedCachedEvents.data = updatedCachedEvents.data.concat(
+      await this.levelDBCacheService.insertEventsCacheBatch({
+        headers: {
```
Amuhar, line 270: do we need to set version?

src/contracts/deposit/integrity-checker/integrity-checker.service.ts
Amuhar: I want to highlight that  in historical front-run check (and check of front-run attempt too) we want to have 'latest' data.

src/contracts/deposit/deposit.service.ts
Amuhar: I just wanted to highlight that we fetch events cache twice in `initialize` method.

src/contracts/deposit/leveldb/leveldb.service.ts
Amuhar: got it.
generateDepositCacheKey is fine

src/contracts/deposit/deposit.service.ts
Amuhar: I think we currently fetch events for every new block we handle in handleNewBlock. We just update the cache for (blockNumber - lag)

src/contracts/deposit/deposit.service.ts
eddort: it's not a big problem because we're pulling events from the local database. It can be optimized, but it won't give much change at startup, because the query to the node takes many times more time
But if we do decide to do that, we'll have to complicate the code, which we don't want to do

src/contracts/deposit/integrity-checker/integrity-checker.service.ts
eddort: we have all the data we need
I just keep `finalized` separate from `latest`.

src/contracts/deposit/deposit.service.ts
eddort: fixed

src/contracts/deposit/deposit.service.ts
eddort: fixed

src/contracts/deposit/deposit.service.spec.ts
eddort: fixed

src/contracts/deposit/deposit.service.spec.ts
eddort: fixed

src/contracts/deposit/deposit-tree/deposit-tree.ts
```diff
@@ -0,0 +1,138 @@
...
+      let height = 0;
+      height < DepositTree.DEPOSIT_CONTRACT_TREE_DEPTH;
+      height++
+    ) {
+      if ((size & 1) == 1) {
+        this.branch[height] = node;
+        return this.branch;
+      }
+
+      node = digest2Bytes32(this.branch[height], node);
```
Amuhar, line 59: look like this.branch[height] can be undefined

Amuhar: maybe here should be smth like 

if (this.branch[height] === undefined) {
      this.branch[height] = this.zeroHashes[height];
 }

src/contracts/deposit/deposit-tree/deposit-tree.ts
```diff
@@ -0,0 +1,138 @@
...
+    depositCount: number,
+  ): Uint8Array[] | undefined {
+    let size = depositCount;
+    for (
+      let height = 0;
+      height < DepositTree.DEPOSIT_CONTRACT_TREE_DEPTH;
+      height++
+    ) {
+      if ((size & 1) == 1) {
+        this.branch[height] = node;
```
Amuhar, line 55: this.branch[height] = digest2Bytes32(this.branch[height], node); ?

src/contracts/deposit/integrity-checker/integrity-checker.service.ts
```diff
@@ -0,0 +1,142 @@
...
+    await this.putEventsToTree(this.finalizedTree, eventsCache);
+    return this.finalizedTree;
+  }
+
+  /**
+   * Inserts a list of latest verified deposit events into a clone of the deposit tree and returns the cloned tree.
+   * @param {VerifiedDepositEvent[]} eventsCache - Array of verified deposit events to be added to the cloned tree.
+   * @returns {Promise<DepositTree>} The cloned and updated deposit tree after adding the events.
+   */
+  public async putLatestEvents(
```
Amuhar, line 45: didnt get idea of having to different functions putFinalizedEvents and putLatestEvents

src/contracts/deposit/deposit-tree/deposit-tree.ts
```diff
@@ -0,0 +1,138 @@
...
+        return this.branch;
+      }
+
+      node = digest2Bytes32(this.branch[height], node);
+
+      size /= 2;
```
Amuhar, line 61: As we check if a number is even or odd with a bitwise operation , we could also use the similar bitwise operation here `size >>= 1`

src/contracts/deposit/deposit-tree/deposit-tree.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+      pubkey: '0xaabbccdd',
+      wc: '0x11223344',
+      amount: '0x0000000000000001', // Little endian of 1
+      signature: '0x55667788',
+    };
+    depositTree.insert(nodeData);
+    expect(depositTree.nodeCount).toBe(1);
+  });
+
+  test('clone should create an exact copy of the tree', () => {
```
Amuhar, line 27: lets add test on odd number of nodes?

src/contracts/deposit/deposit-tree/deposit-tree.ts
```diff
@@ -0,0 +1,138 @@
...
+      let height = 0;
+      height < DepositTree.DEPOSIT_CONTRACT_TREE_DEPTH;
+      height++
+    ) {
+      if ((size & 1) == 1) {
+        node = digest2Bytes32(this.branch[height], node);
+      } else {
+        node = digest2Bytes32(node, this.zeroHashes[height]);
+      }
+      size /= 2;
```
Amuhar, line 101: will it work properly if size is odd? or we will get here an odd number

eddort: Using `size /= 2` is not a mistake. In JavaScript, when performing bitwise operations like `& 1`, floating-point numbers are implicitly converted to integers, discarding the fractional part. This ensures the algorithm works correctly and matches the logic of a Solidity smart contract. Solidity does not have floating-point numbers, and all division is performed as integer division, rounding down the result.

src/contracts/deposit/deposit-tree/deposit-tree.spec.ts
eddort: added

src/contracts/deposit/deposit-tree/deposit-tree.ts
eddort: I added a comment to the code describing why I did it this way

Amuhar (approved):

PR #201, opened 2024-05-08
Test the deposits' pause with Anvil in version 3

Amuhar:


test/helpers/anvil-fork.ts
```diff
@@ -0,0 +1,48 @@
...
+      '-f',
+      this.forkUrl,
+      '-p',
+      this.port,
+      forkBlockCommand,
+      '--block-time',
+      '12',
+      '--auto-impersonate',
+    ]);
+
```
eddort, line 38: While developing the library to run `anvil` in Docker (https://github.com/eddort/anvild), I encountered an issue where when `anvil` is started in Docker with fork mode enabled, an error can occur indicating that the process is already running, but the `anvil` node has not yet synchronized. To solve this problem, I used code from another library (https://github.com/wevm/anvil.js).

Below is a solution that helps with this problem:

```ts
export const waitForServerStdout = (stream: NodeJS.ReadWriteStream) => {
  return new Promise((res) => {
    stream.on("data", (data: any) => {
      const output = data.toString();
      if (output.includes("Listening")) {
        res(true);
      }
    });
  });
};
```

PR #206, opened 2024-05-27
Duplicates improvement

Amuhar:
Cache SigningKeyAdd in leveldb;
improve duplicate identification algorithm


src/guardian/duplicates/keys-duplication-checker.service.ts
```diff
@@ -0,0 +1,180 @@
...
+    const { events } =
+      await this.signingKeyEventsCacheService.getUpdatedSigningKeyEvents(
+        key,
+        blockData.blockNumber,
+        blockData.blockHash,
+      );
+    return events;
+  }
+
+  private findOriginalEvent(events: any[]): any {
```
eddort, line 137: in this file in several places `events` are listed as `any`, let's fix it, shall we?

src/guardian/duplicates/keys-duplication-checker.service.ts
```diff
@@ -0,0 +1,180 @@
...
+   *
+   * Duplicates Search Algorithm:
+   * 1. If there are duplicates within one operator, the key with the lowest index is considered the original, and the others are considered duplicates.
+   * 2. If there are duplicates between different operators, check if a deposited key exists in the duplicates list; all others are considered duplicates.
+   * 3. If there is no deposited key, check the SigningKeyAdded events for operators.
+   * 4. Sort events by block number and logIndex. The earliest event is considered the original, and the others are marked as duplicates.
+   *
+   * What if there is no event for the key? This might indicate that the node returned an incorrect answer, the module didn't implement SigningKeyAdded, or there is a mistake in the cache algorithm. Should we throw an error?
+   *
+   * TODO: If there is no event for the key, should it be considered added in block 0? (Should this be done only for the curated module?)
```
eddort, line 26: Based on the current logic, there should be no keys without events
As far as I remember it could be if new events with indexes were used, but these events were there from the beginning.

src/guardian/duplicates/keys-duplication-checker.service.ts
```diff
@@ -0,0 +1,180 @@
...
+   */
+  async getDuplicatedKeys(
```
eddort, line 28: I propose to divide this method into several smaller methods

src/guardian/duplicates/keys-duplication-checker.service.ts
Amuhar: it is more about case of error in storage.

src/contracts/signing-key-events-cache/signing-key-events-cache.service.ts
```diff
@@ -0,0 +1,321 @@
...
+   * @returns event group
+   */
+  public async getCachedEvents(): Promise<SigningKeyEventsCache> {
+    const { headers, data } = await this.levelDBCacheService.getEventsCache();
+
+    // default values is startBlock: 0, endBlock: 0
+    const deploymentBlock = await this.getDeploymentBlockByNetwork();
+
+    return {
+      headers: {
```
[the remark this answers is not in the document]
Amuhar: we could update list, instead of delete. even if it take 3 minutes to fetch events maybe better not to lose cache

src/contracts/deposit/deposit.service.ts
```diff
@@ -329,6 +329,7 @@ export class DepositService {
     if (!isCacheValid) process.exit(1);
 
     const firstNotCachedBlock = cachedEvents.headers.endBlock + 1;
+    // TODO: if blockNumber == cachedEvents.headers.endBlock, than firstNotCachedBlock > endBlock
```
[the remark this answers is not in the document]
Amuhar: i think it should not happen

src/contracts/signing-key-events-cache/signing-key-events-cache.service.ts
```diff
@@ -0,0 +1,395 @@
...
+      return false;
+    }
+
+    const currentModules = await this.getStakingModules();
+    const isDifferentLength = currentModules.length !== previousModules.length;
+    const hasNewModules = currentModules.some(
+      (module) => !previousModules.includes(module),
+    );
+
+    const wasUpdated = isDifferentLength || hasNewModules;
```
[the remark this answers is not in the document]
Amuhar: I see that it might not be obvious, and it is better to write it the way you did. However, I wrote a test to cover all cases, so you might find a case that it doesn't cover.

I will change the code as you suggested too

eddort (approved):

PR #210, opened 2024-06-17
fix: parallel call unvetSignKeysChunk

eddort:


src/guardian/unvetting/unvetting.service.ts
```diff
@@ -78,28 +78,33 @@ export class UnvettingService {
           vettedKeysByOperator,
         );
 
-        await this.securityService
-          .unvetSigningKeys(
+        const results = await Promise.allSettled([
+          this.securityService.unvetSigningKeys(
```
Amuhar, line 82: if we want concurrently send transaction and message in broker here, maybe we want the same strategy for pause, could you  check it too ?

Amuhar (approved):

PR #216, opened 2024-08-07
Fix bug in keys add events caching

Amuhar:


src/contracts/signing-key-events-cache/leveldb/leveldb.service.ts
```diff
@@ -199,6 +205,28 @@ export class LevelDBService {
     await this.db.batch(ops);
   }
 
+  private validateHeader(
```
eddort, line 208: Perhaps we should think about some kind of library for schema validation? Like zod?

src/contracts/signing-key-events-cache/signing-key-events-cache.service.ts
```diff
@@ -70,12 +76,18 @@ export class SigningKeyEventsCacheService {
...
     const firstNotCachedBlock = initialCache.headers.endBlock + 1;
     const toBlock = latestBlock - SIGNING_KEYS_EVENTS_CACHE_LAG_BLOCKS;
 
     const totalEventsCount = initialCache.data.length;
     let newEventsCount = 0;
 
+    const stakingModulesContracts =
+      await this.repositoryService.getCachedStakingModulesContracts();
+    const stakingModulesAddresses = Object.keys(stakingModulesContracts);
+
```
eddort, line 90: Should we replace this call with `getStakingModules`? As I see it, this function does the same job

eddort: And should we re-check the list of modules in each `getUpdatedSigningKeyEvents` call?

eddort:
Great job! Thank you!

src/contracts/signing-key-events-cache/leveldb/leveldb.service.ts
Amuhar: I am not sure that we should install a library for just one check. If this check is enough, I will not delegate it to a library.

src/contracts/signing-key-events-cache/signing-key-events-cache.service.ts
Amuhar: I don't think so, as we check it in handleNewBlock and use getUpdatedSigningKeyEvents in the same iteration to search for duplicates.

eddort (approved):

PR #219, opened 2024-08-13
Return all Instances of keys with invalid signatures, including duplicates

Amuhar:


eddort (approved):
Awesome job! Thank you!

PR #232, opened 2024-08-29
Historical frontrun fix

Amuhar:


test/front-run-v3.e2e-spec.ts
```diff
@@ -611,4 +613,134 @@ describe('ganache e2e tests', () => {
     },
     TESTS_TIMEOUT,
   );
+
+  test(
+    'should not trigger pause for front-run attempt with non-Lido WC and Lido WC deposits when key is unused',
```
eddort, line 618: I would only suggest to keep this test + add another test where it would be paused (those key used: true)

eddort (approved):

PR #234, opened 2024-09-01
Fix duplicates inconsistency with doc

Amuhar:


src/guardian/duplicates/keys-duplication-checker.service.ts
```diff
@@ -167,37 +235,84 @@ export class KeysDuplicationCheckerService {
...
-        (prev.blockNumber === curr.blockNumber && prev.logIndex < curr.logIndex)
-          ? prev
-          : curr,
-      events[0],
+  private getOperatorsWithoutEvents(
+    operators: string[],
+    events: SigningKeyEvent[],
+  ): string[] {
+    const eventOperators = new Set(
+      events.map((event) => `${event.moduleAddress}-${event.operatorIndex}`),
```
eddort, line 243: small improvement: put `${event.moduleAddress}-${event.operatorIndex}` in a separate method

src/guardian/duplicates/keys-duplication-checker.service.ts
```diff
@@ -77,35 +51,107 @@ export class KeysDuplicationCheckerService {
...
+      suspectedDuplicateKeys,
+      operators,
+      blockData,
     );
   }
 
-  private extractOperators(occurrences: RegistryKey[]): Set<string> {
-    return new Set(
-      occurrences.map((key) => `${key.moduleAddress}-${key.operatorIndex}`),
+  private getOperators(keys: RegistryKey[]): string[] {
```
eddort, line 102: method does not accurately reflect the essence: this method returns idempotency keys, I would suggest to reflect this in the name and in all arguments of functions to which it is passed.

src/guardian/duplicates/keys-duplication-checker.service.ts
```diff
@@ -167,37 +235,84 @@ export class KeysDuplicationCheckerService {
...
-  ): string[] {
-    const eventOperators = new Set(
-      events.map((event) => `${event.moduleAddress}-${event.operatorIndex}`),
+  private findEarliestEvents(events: SigningKeyEvent[]): SigningKeyEvent[] {
+    if (events.length <= 1) return events;
+
+    const { blockEvents } = events.reduce(
+      ({ earliestBlockNumber, blockEvents }, currEvent) => {
+        if (earliestBlockNumber === currEvent.blockNumber) {
+          blockEvents.push(currEvent);
```
eddort, line 281: This pattern can lead to infinite loops in case of an error, it is better to avoid it. I suggested to reconsider the solution of this method in favour of Math.min and filter

eddort (approved):

PR #235, opened 2024-09-02
Small refactoring block-guard service

Amuhar:


src/guardian/guardian.service.spec.ts
```diff
@@ -104,7 +103,7 @@ describe('GuardianService', () => {
     }));
 
     const getBlockGuardServiceMock = jest
-      .spyOn(blockGuardService, 'isNeedToProcessNewState')
+      .spyOn(guardianService, 'isNeedToProcessNewState')
```
eddort, line 106: 'blockGuardService' is defined but never used

src/guardian/guardian.service.ts
```diff
@@ -412,4 +412,32 @@ export class GuardianService implements OnModuleInit {
       stakingModuleData.isModuleDepositsPaused
     );
   }
+
+  public isNeedToProcessNewState(newMeta: {
```
eddort, line 416: Could you tell me why you decided to move this method?

src/guardian/guardian.service.spec.ts
Amuhar: changed in fix/val-1216-gurdian-service-impr

src/guardian/guardian.service.ts
Amuhar: https://github.com/lidofinance/lido-council-daemon/pull/220/files#r1732405735 

this cache belong more to guardian.service

eddort: I have already moved this method, it was here before, I think it is more convenient when there are no auxiliary methods in the main entity. This method looks good in the Blocks service as well

Amuhar: With this method, we check whether we have already verified the keys on this block, whether we need to process a new state, sign deposit messages, and so on. BlockGuardService does not handle these tasks. Instead, BlockGuardService (which will be renamed to BlockDataCollector in the future) contains methods to fetch and collect data from contracts. It also doesn't know anything about the keys API.

GuardianService is the main module responsible for running methods to check keys on this block, process a new state, sign deposit messages, and so on. It should maintain the state necessary to determine whether we need to process a new state.

eddort (approved):

infloop (approved):

PR #237, opened 2024-09-03
Improve validation algorithm

Amuhar:


src/guardian/keys-validation/keys-validation.service.ts
```diff
@@ -46,91 +43,92 @@ export class KeysValidationService {
...
+      cachedInvalidKeyList: RegistryKey[];
+      uncachedDepositKeyList: DepositKey[];
     }>(
-      (acc, depositData) => {
-        const cacheResult = this.getCachedDepositData(depositData);
-
-        if (cacheResult === false || cacheResult === true) {
-          acc.cachedDepositData.push([depositData, cacheResult]);
+      (acc, key) => {
+        const depositKey = {
```
eddort, line 109: The main difference here is that we take as part of the search key withdrawalCredentials and genesisForkVersion?
Is there anything else to look out for when reading?

src/guardian/keys-validation/keys-validation.spec.ts
```diff
@@ -47,95 +49,67 @@ describe('KeysValidationService', () => {
     jest.spyOn(loggerService, 'log').mockImplementation(() => undefined);
   });
 
-  it('should find and return invalid keys from the provided list', async () => {
```
eddort, line 50: don't we need this test case?

src/guardian/staking-module-guard/staking-module-guard.spec.ts
```diff
@@ -369,7 +365,6 @@ describe('StakingModuleGuardService', () => {
         nonce: 1,
         blockNumber: 100,
         lastChangedBlockHash: 'hash',
-        invalidKeysFound: false,
```
eddort, line 372: What this variable means? unfortunately I don't see these changes in diff.

src/guardian/keys-validation/keys-validation.service.ts
```diff
@@ -46,91 +43,92 @@ export class KeysValidationService {
...
+    });
+
+    const validatedDepositKeyList: [DepositKey & Key, boolean][] =
+      await this.keyValidator.validateKeys<DepositKey>(uncachedDepositKeyList);
 
+    this.updateCache(validatedDepositKeyList);
+
+    const invalidKeys = this.filterInvalidKeys(validatedDepositKeyList);
+
+    return [...cachedInvalidKeyList, ...invalidKeys];
```
infloop, line 67: Better to use `concat` here to join arrays, because `...` can cause stack overflow on big arrays.

src/guardian/keys-validation/keys-validation.service.ts
Amuhar: withdrawalCredentials and genesisForkVersion as part of key we used before too
we use as key whole deposit data for caching 
But in this pr we use 
RegistryKey & {
  withdrawalCredentials: WithdrawalCredentialsBuffer;
  genesisForkVersion: Buffer;
}; type for validation

to get back list of RegistryKey & {
  withdrawalCredentials: WithdrawalCredentialsBuffer;
  genesisForkVersion: Buffer;
};

and just map result to RegistryKey

Amuhar: prev we needed to make a search across result of validation to find all matching RegistryKey

src/guardian/keys-validation/keys-validation.spec.ts
Amuhar: list of tests there revised
we still have test that validate without  use of cahce , after with use of cache and than if deposited data were changed valdiate again without cche

PR #244, opened 2024-09-16
Feat/signing keys registry

eddort:


src/guardian/guardian.service.ts
```diff
@@ -108,8 +107,8 @@ export class GuardianService implements OnModuleInit {
 
         // The event cache is stored with an N block lag to avoid caching data from uncle blocks
         // so we don't worry about blockHash here
-        // TODO: rewrite signingKeyEventsCacheService
-        await this.signingKeyEventsCacheService.updateEventsCache(
+        // TODO: rewrite SigningKeysRegistryService
```
Amuhar, line 110: remove TODO

src/guardian/duplicates/keys-duplication-checker.service.ts
```diff
@@ -241,13 +242,20 @@ export class KeysDuplicationCheckerService {
...
-      await this.signingKeyEventsCacheService.getUpdatedSigningKeyEvents(
+    const eventsGroup =
+      await this.signingKeysRegistryService.getUpdatedSigningKeyEvents(
         key,
         blockData.blockNumber,
         blockData.blockHash,
       );
-    return events;
+
+    if (!eventsGroup.isValid) {
```
Amuhar, line 252: it will break whole keys check
instead based on duplicates logic if we are not sure that events cache correct we should put this events in a list of unresolvedDuplicates

src/guardian/guardian.service.ts
eddort: fixed

src/guardian/duplicates/keys-duplication-checker.service.ts
eddort: fixed

src/contracts/signing-keys-registry/signing-keys-registry.service.ts
```diff
@@ -384,6 +314,7 @@ export class SigningKeyEventsCacheService {
       stakingModulesAddresses: cachedEvents.headers.stakingModulesAddresses,
       startBlock: cachedEvents.headers.startBlock,
       endBlock,
+      isValid,
```
Amuhar, line 317: i dont think we should here the same as in deposit events cache
these two caches solve diff problems
we want to make pause even if events list is not valid
but we dont want to make unvetting if this chache is not valid

I see here two options:

return in this method only valid part of cache and don't  set isValid in header at all
if for duplicates identification we don't have enough data, we will process them as unresolved, as we already do
pluses: we will know for what keys we don't have events in cache, it will help partially resolve duplicates
unresolved list has the same meaning as isValid

Amuhar (approved):

PR #257, opened 2024-10-05
fix: leveldb cache perfomance

eddort:


avsetsin (approved):
I think it's fine to merge to develop and deploy it like this to the testnet to check for performance improvements, but would recommend refactoring the code to split responsibility between services. In the current implementation, the `DepositsRegistryStoreService` is responsible for both caching to memory and caching to DB. Consider explicitly splitting these roles between services

PR #280, opened 2025-03-11
feat: enhance event fetching

eddort:


src/contracts/data-bus/data-bus.client.ts
```diff
@@ -29,10 +32,21 @@ export class DataBusClient {
   }
 
   async sendTransaction(eventId: string, dataBytes: string) {
+    const estimatedGas = await this.dataBus.estimateGas.sendMessage(
+      eventId,
+      dataBytes,
+    );
+
+    const gasLimit = estimatedGas.mul(DATA_BUS_GAS_MULTIPLIER).div(100);
```
F4ever, line 40: Optional:
in depositor bot we have sanity check `CONTRACT_GAS_LIMIT` to be sure we won't spent more fee than we expect

src/bls/bls.service.ts
```diff
@@ -64,9 +66,17 @@ export class BlsService implements OnModuleInit {
 
       return blst.verify(signingRoot, blsPublicKey, blsSignature);
```
F4ever, line 67: Let's cover with try catch only part with `blst.verify`

PR #281, opened 2025-03-12
Fix/holesky events fetching

eddort:


src/contracts/deposits-registry/deposits-registry.constants.ts
```diff
@@ -8,7 +8,12 @@ export const DEPLOYMENT_BLOCK_NETWORK: {
   [CHAINS.Holesky]: 0,
 };
 
-export const DEPOSIT_EVENTS_STEP = 10_000;
+export const DEPOSIT_EVENTS_STEP: {
```
F4ever, line 11: Should we specify default value somehow here?

src/contracts/deposits-registry/deposits-registry.service.ts
```diff
@@ -259,6 +273,14 @@ export class DepositRegistryService {
     };
   }
 
+  public async getDepositEventStep(): Promise<number> {
+    const chainId = await this.providerService.getChainId();
+    const step = DEPOSIT_EVENTS_STEP[chainId];
+    if (step == null) throw new Error(`Chain ${chainId} is not supported`);
```
F4ever, line 279: Maybe define default value for other networks?

F4ever (approved):
There is no test for `getDepositEventStep` method.

Rest LGTM

Amuhar (approved):

PR #290, opened 2025-05-21
Next release

eddort:
* E2E tests
* Update to Node.js 18
* Dependency updates (bls)
* Removed legacy logic for handling two versions of the DSM contract
* Increased Node.js memory limit to 8GB
* Changed data-bus configuration (timeout increased from 10 seconds to 30 seconds)
* Proper error handling in case the data-bus provider is unavailable


src/transport/data-bus/data-bus.transport.ts
```diff
@@ -23,14 +24,30 @@ export class DataBusTransport implements TransportInterface, OnModuleInit {
...
   public async initialize() {
+    try {
+      await this.connect();
+    } catch (error) {
+      this.logger.warn('Error initializing DataBusTransport', error);
+    }
+  }
+
+  private async connect() {
+    if (this.isDataBusConnected) {
```
Amuhar, line 35: if it once set to `true`, it will not try to connect again
is it possible ?

Amuhar: was discussed that this check need only during application start

Amuhar (approved):

Ziars (approved):

PR #314, opened 2026-02-03
feat: switch provider only on error

eddort:


infloop (dismissed):

infloop (dismissed):

infloop (dismissed):

src/provider/provider.constants.ts
```diff
@@ -1 +1,3 @@
-export const MAIN_PROVIDER_REQUEST_TIMEOUT = 10 * 60 * 1000; // 10 minutes
+export const MAIN_PROVIDER_REQUEST_TIMEOUT = 15 * 60 * 1000; // 15 minutes
```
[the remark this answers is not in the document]
eddort: will fix later

src/provider/provider.constants.ts
```diff
@@ -1 +1,3 @@
...
+// Don't reset provider URL selection, only on error
+export const PROVIDER_RESET_INTERVAL = 60 * 60 * 1000 * 24 * 10; // 10 days
```
F4ever:
```suggestion
export const MAIN_PROVIDER_REQUEST_TIMEOUT = 5 * 60 * 1000; // 5 minutes
// Don't reset provider URL selection, only on error
export const PROVIDER_RESET_INTERVAL = 60 * 60 * 1000 * 24; // 1 day
```

src/provider/main-provider.module.ts
```diff
@@ -22,7 +25,10 @@ export class MainProviderModule {
             urls: config.PROVIDERS_URLS ?? [config.RPC_URL],
             network: config.CHAIN_ID,
             instanceLabel: 'EL1',
+            maxRetries: 1,
```
F4ever, line 28: Maybe 2?

```suggestion
            maxRetries: 2,
```

src/contracts/deposits-registry/deposits-registry.constants.ts
```diff
@@ -12,7 +12,7 @@ export const DEPLOYMENT_BLOCK_NETWORK: {
 export const DEPOSIT_EVENTS_STEP: {
   [key in CHAINS]?: number;
 } = {
-  [CHAINS.Mainnet]: 10_000,
+  [CHAINS.Mainnet]: 500,
```
F4ever: It will increase rpc consumtion x20 and will increase loading time.
Is this required?

src/contracts/deposits-registry/fetcher/fetcher.service.ts
```diff
@@ -48,14 +48,58 @@ export class DepositsRegistryFetcherService {
...
+
+    this.logger.log('fetchEvents: starting queryFilter', {
+      startBlock,
+      endBlock,
+    });
+    const queryStartTime = Date.now();
     const filter = contract.filters.DepositEvent();
     const rawEvents = await contract.queryFilter(filter, startBlock, endBlock);
-    const events = rawEvents.map((rawEvent) => {
+    const queryDurationMs = Date.now() - queryStartTime;
```
F4ever, line 64: We have RPC metrics for this. Do we need addititonal calculations here?

F4ever (changes requested):
Review

src/utils/fetch-events-utils.ts
```diff
@@ -48,25 +48,43 @@ export async function fetchEventsFallOver<
   const stack: RangeTask[] = [{ start: startBlock, end: endBlock }];
 
   let task: RangeTask | undefined;
+  let fetchAttempt = 0;
+
   while ((task = stack.pop()) !== undefined) {
+    fetchAttempt++;
```
F4ever: it's not fetchAttempt. Maybe `requestCount`?

src/utils/fetch-events-utils.ts
```diff
@@ -75,9 +93,11 @@ export async function fetchEventsFallOver<
       stack.push(right);
       stack.push(left);
 
-      logger?.debug?.('Splitting range', {
+      logger?.warn?.('Splitting range due to provider failure', {
         start: task.start,
         end: task.end,
+        newRanges: { left, right },
+        durationMs: fetchDuration,
```
F4ever, line 100: there is no fetch? Just splitting task

eddort: no, it exists but in next iteration because of while loop

src/contracts/deposits-registry/fetcher/fetcher.service.ts
eddort: yep, we need this for double check of fallback provider data

src/provider/main-provider.module.ts
eddort: we need `1` here because of splitting task algorithm

src/provider/provider.constants.ts
eddort: i'll rollback to 10 min, but `MAIN_PROVIDER_REQUEST_TIMEOUT` i'ts like "noop" in this point

src/contracts/deposits-registry/deposits-registry.constants.ts
eddort: i'll rollback it

infloop (dismissed):

F4ever (dismissed):

hweawer (approved):

PR #316, opened 2026-03-09
Support 0x02 withdrawal credentials & e2e test on mainnet

Amuhar:


src/guardian/staking-module-guard/staking-module-guard.service.ts
```diff
@@ -214,23 +218,27 @@ export class StakingModuleGuardService {
...
   /**
    * Checks if Lido deposits have been front-ran in the past based on historical deposit data.
-   * This method does not account for WC rotation as historical deposits were manually checked.
    *
    * @param depositedEvents - A group of historical deposit events.
-   * @param lidoWC - The withdrawal credential associated with Lido.
+   * @param lidoWCList - All valid Lido withdrawal credentials (0x01 and 0x02 variants).
    * @returns True if front-running was detected at any point in the past; false if no front-running occurred.
    */
   public async getHistoricalFrontRun(
```
dry914: `getHistoricalFrontRun` and its 6 private helpers (`getDepositsWithLidoWC, getEarliestDepositsMap, getNonLidoDuplicatedDeposits, getValidNonLidoDuplicatedDeposits, getFrontRun, getKeysDepositedByLido`) are no longer called - replaced by `checkValidatorsHasWrongWC`. ~170 lines of dead code + commented-out call in `block-data-collector.service.ts:77`. Should be removed.

src/guardian/staking-module-guard/staking-module-guard.service.ts
```diff
@@ -263,6 +271,80 @@ export class StakingModuleGuardService {
...
+        {
+          wrongWCTypeLidoDeposits,
+        },
+      );
+    }
+
+    return hasFrontRunning || hasWrongWCType;
+  }
+
   public async alreadyPausedDeposits(blockData: BlockData, version: number) {
```
dry914: `StakingModuleGuardService.alreadyPausedDeposits()` is dead code - replaced by a private method in `BlockDataCollectorService`. Should be removed.

src/contracts/staking-router/staking-router.service.ts
dry914: Remove await for consistency?

src/guardian/staking-module-guard/staking-module-guard.service.ts
```diff
@@ -263,6 +271,80 @@ export class StakingModuleGuardService {
...
   }
 
+  /**
+   * Check if all deposited keys has withdrawal credentials corresponding to module
+   *
+   * @param depositedEvents All verified deposit contract events
+   * @param usedKeys lido keys
+   * @param moduleAddressToWC Map of addresses match withdrawal credentials
+   */
+  public async checkValidatorsHasWrongWC(
```
eddort: The method is currently called `checkValidatorsHasWrongWC`, but internally it performs two checks:
- a front-run check
- a wrong WC check

It then returns a single flag, and based on that flag we pause deposits.

This creates an important monitoring problem: apart from the service logs, we have no way to understand why exactly we are pausing deposits. The method also mixes responsibilities.

I suggest splitting this method into two parts, where the two predicates are checked separately, and returning flags and structures that we can then pass into metrics as labels.

src/guardian/guardian.service.ts
```diff
@@ -370,21 +385,27 @@ export class GuardianService implements OnModuleInit {
...
     alreadyPausedDeposits: boolean,
     stakingModuleId: number,
   ): boolean {
     const keysForUnvetting = stakingModuleData.invalidKeys.concat(
       stakingModuleData.frontRunKeys,
       stakingModuleData.duplicatedKeys,
+      stakingModuleData.crossTypeKeys,
     );
 
+    // TODO: no need to check (hasFrontRunning || hasWrongWCType)
```
eddort: The comment is not relevant, should we delete it?

eddort (approved):
There are still some lint problems, let's fix them?

PR #318, opened 2026-03-11
feat(orc-635): empty deposits council deprecation hotfix

chasingrainbows:


src/staking-module-data-collector/staking-module-data-collector.service.ts
```diff
@@ -39,16 +39,27 @@ export class StakingModuleDataCollectorService {
...
+
+    const depositableEther =
+      await this.stakingRouterService.getDepositableEther(blockTag);
+
     return await Promise.all(
       stakingModules.map(async (stakingModule) => {
+        const maxDepositsCount =
+          await this.stakingRouterService.getStakingModuleMaxDepositsCount(
+            stakingModule.id,
+            depositableEther,
```
F4ever: Let's check allocation with 32 ETH only always. So we sign deposit into one most prioritized SM

src/staking-module-data-collector/staking-module-data-collector.service.ts
```diff
@@ -39,16 +39,27 @@ export class StakingModuleDataCollectorService {
...
         return {
           isModuleDepositsPaused:
             await this.stakingRouterService.isModuleDepositsPaused(
               stakingModule.id,
-              {
-                blockHash: meta.blockHash,
-              },
+              blockTag,
             ),
+          hasDepositsAllocation: maxDepositsCount > 0,
```
F4ever, line 72: What we do if all modules has `hasDepositsAllocation` == false?

chasingrainbows: No deposit messages are signed, Promise.all resolves and the function completes. Next block the daemon re-evaluates. No action needed from our side, no stuck state. The daemon will pick up deposits again as soon as allocation becomes available.

src/staking-module-data-collector/staking-module-data-collector.service.ts
chasingrainbows: done

F4ever (approved):

PR #330, opened 2026-06-23
[EPIC] Council - SRv3

F4ever:


src/contracts/signing-keys-registry/store/store.service.ts
```diff
@@ -6,13 +6,26 @@ import { SimpleFallbackJsonRpcBatchProvider } from '@lido-nestjs/execution';
...
+  METRIC_SIGNING_KEYS_CACHE_BYTES,
+  METRIC_SIGNING_KEYS_CACHE_COUNT,
+} from 'common/prometheus';
+import { Gauge } from 'prom-client';
 
 @Injectable()
 export class SigningKeysStoreService {
   private db!: Level<string, string>;
+  private cacheSizeBytes = 0;
+  private cacheEventCount = 0;
```
tamtamchik, line 20: 🧠  looks almost the same as`store.service.ts`. Move it to a common class?

test/duplicates-v3.e2e-spec.ts
```diff
@@ -176,8 +180,13 @@ describe('Duplicates e2e tests', () => {
 
     const shouldReceiveDeposits =
       getModuleIssuesCount(moduleState) === 0 &&
-      !moduleState.isModuleDepositsPaused &&
-      moduleState.hasDepositsAllocation;
+      !moduleState.isModuleDepositsPaused;
+
+    console.log(
```
tamtamchik, line 185: Remove `console.log`?

test/duplicates-v3.e2e-spec.ts
```diff
@@ -201,18 +210,30 @@ describe('Duplicates e2e tests', () => {
...
       }),
       stakingModuleDataCollectorService.collectStakingModuleData({
         stakingModules,
         meta: elBlockSnapshot,
         lidoKeys,
+        moduleWCMap,
       }),
     ]);
 
+    console.log('stakingModulesData: ', stakingModulesData);
```
tamtamchik, line 235: Remove?

test/helpers/dsm.ts
```diff
@@ -98,31 +116,23 @@ export async function deposit(moduleId: number, depositCount = 1) {
...
   await accountImpersonate(agent);
-  await setBalance(dsm, 100);
+  await accountImpersonate(daoAddress);
   await setBalance(agent, 100);
-  const signer = testSetupProvider.getSigner(dsm);
 
-  const lido = LidoAbi__factory.connect(lidoAddress, signer);
+  const agentSigner = testSetupProvider.getSigner(agent);
+  const lido = LidoAbi__factory.connect(lidoAddress, agentSigner);
+  const lidoAgentSigner = LidoAbi__factory.connect(lidoAddress, agentSigner);
```
tamtamchik: Why both `lido` and `lidoAgentSigner` use `lidoAddress`?

src/contracts/security/security.service.ts
```diff
@@ -374,13 +374,19 @@ export class SecurityService {
 
     const currentVersion = version.toNumber();
 
-    if (currentVersion !== DSM_CONTRACT_SUPPORTED_VERSION) {
-      this.logger.warn(`Deprecated DSM contract version found: ${version}`, {
+    const isSupportedVersion = DSM_CONTRACT_SUPPORTED_VERSIONS.some(
+      (supportedVersion) => supportedVersion === currentVersion,
+    );
```
tamtamchik, line 379:
```suggestion
    const isSupportedVersion = DSM_CONTRACT_SUPPORTED_VERSIONS.includes(currentVersion);
```

src/guardian/deposit-allocation/dsm-deposit-allocation-adapter.service.ts
```diff
@@ -0,0 +1,56 @@
...
+    const maxDepositsCount =
+      await this.stakingRouterService.getStakingModuleMaxDepositsCount(
+        stakingModuleData.stakingModuleId,
+        allocationCheckValue,
+        blockTag,
+      );
+
+    const hasDepositsAllocation = maxDepositsCount > 0;
+
+    return !hasDepositsAllocation;
```
tamtamchik, line 54: Not sure about `<=` but this does reads easier
```suggestion
    return maxDepositsCount <= 0;
```

src/common/rpc-metrics/src/rpc-metrics.service.ts
```diff
@@ -0,0 +1,354 @@
...
+    }
+  }
+
+  private extractHostname(domain: string): string {
+    if (!domain.startsWith('http')) {
+      return domain;
+    }
+
+    const parsed = new URL(domain);
+    return parsed.port ? `${parsed.hostname}:${parsed.port}` : parsed.hostname;
```
tamtamchik, line 302: I think this should work
```suggestion
    return new URL(domain).host;
```

src/guardian/unvetting/unvetting.service.ts
```diff
@@ -149,8 +154,8 @@ export class UnvettingService {
     );
   }
 
-  private async getMaxOperatorsPerUnvetting() {
-    return await this.securityService.getMaxOperatorsPerUnvetting();
+  private async getMaxOperatorsPerUnvetting(blockTag: BlockTag) {
```
tamtamchik, line 157: Why do we need it as a separate method?

src/guardian/staking-module-guard/staking-module-guard.service.ts
```diff
@@ -59,222 +60,120 @@ export class StakingModuleGuardService {
...
-   * @param nonLidoDuplicatedDeposits - An array of duplicated deposit events with non-Lido withdrawal credentials
-   * @returns  An array of valid duplicated deposit events.
+   * Check if any used lido key has earliest deposit with non-lido withdrawal credentials (theft)
    */
-  private getValidNonLidoDuplicatedDeposits(
-    nonLidoDuplicatedDeposits: VerifiedDepositEvent[],
-  ): VerifiedDepositEvent[] {
-    const validNonLidoDuplicatedDeposits = nonLidoDuplicatedDeposits.filter(
-      (event) => event.valid,
+  public checkHistoricalFrontRun(
```
tamtamchik, line 131: `checkHistoricalFrontRun` and `checkWrongWCType` recalculate the same thing and are always called together. Combine them into a single method?

src/provider/data-bus-provider.module.ts
```diff
@@ -12,11 +12,13 @@ import { PROVIDER_RESET_INTERVAL } from './provider.constants';
 
 export const DATA_BUS_PROVIDER_TOKEN = 'DATA_BUS_PROVIDER';
 
+@Global()
 @Module({})
 export class DataBusProviderModule {
   static forRootAsync(): DynamicModule {
     return {
       module: DataBusProviderModule,
+      global: true,
```
tamtamchik, line 21: Already `@Global()`?

src/provider/test-provider.module.ts
```diff
@@ -14,7 +14,7 @@ export class TestProviderModule {
             // Use new array-based config with fallback to localhost for tests
             urls: config?.PROVIDERS_URLS || ['http://127.0.0.1:8545'],
             // Use chain ID config with fallback to CHAIN_ID env var for tests
-            network: 560048,
+            network: Number(process.env.CHAIN_ID) || 1,
```
tamtamchik, line 17: Is 1 by default safu? Not sure.

test/jest-e2e.setup.ts
```diff
@@ -0,0 +1,3 @@
+if (process.env.E2E_SILENT_LOGS === undefined) {
+  process.env.E2E_SILENT_LOGS = 'true';
+}
```
tamtamchik, line 3:
```suggestion
process.env.E2E_SILENT_LOGS ??= 'true';
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "lido-council-daemon",
-  "version": "3.6.2",
+  "version": "4.0.2",
```
tamtamchik: Why not 4.0.0?
Also version should be reflected in lock file

src/contracts/signing-keys-registry/store/store.service.ts
eddort: I’d keep these stores separate for now. They look similar at the LevelDB level, but their domain semantics are different: key structure, serialization/parsing, cache lifecycle, metrics, and consistency handling differ quite a bit.

Extracting a common base class here would mostly move a few infrastructural lines behind hooks/generics and make both modules harder to read. `deposits-registry` is also the more critical path, so I’d prefer not to couple it to a shared abstraction unless we see a third similar store or repeated maintenance pressure that clearly justifies it.

package.json
eddort: The version from package.json is used as a contract for messages: we use it to understand which version each holder is running. It is 4.0.2, not 4.0.0, because we made two testnet releases, and that needed to be reflected in this file for holder version control.
> Also version should be reflected in lock file

yarn.lock does not store the version of the project itself.

src/provider/test-provider.module.ts
eddort: It is safe because it is test provider for e2e

src/guardian/staking-module-guard/staking-module-guard.service.ts
eddort: We split it up for readability, but now it really does seem better to merge it. However, I’d suggest doing that separately, as a performance-related change.

src/contracts/signing-keys-registry/store/store.service.ts
tamtamchik: Got it, agree. You can mark this one as resolved.

src/guardian/staking-module-guard/staking-module-guard.service.ts
tamtamchik: Agree, not a blocker. Thread be resolved.

package.json
tamtamchik: Thread can be resolved also.

hweawer (dismissed):

test/helpers/dsm.ts
eddort: fixed

hweawer (approved):

PR #332, opened 2026-07-02
Consideration legacy wc in historical front-run

Amuhar:
**Fix historical front-run false positive for legacy withdrawal credentials**

Some mainnet Curated module keys were deposited with the 0x00 BLS withdrawal credentials before the WC was rotated to 0x01. The historical front-run check only compared against the current on-chain WC, so those legitimate legacy deposits were flagged as theft and could pause the module. We now keep a per-network/per-module list of legacy WCs (mainnet module 1 → 0x0096…125e) and treat them as valid for used keys. Vetted-unused keys are unchanged (still unvetted on mismatch).

Also split the front-run / wrong-WC-type metrics into two methods so they no longer overwrite each other.

### DIFF
--- a/src/guardian/staking-module-guard/staking-module-guard.service.ts
+++ b/src/guardian/staking-module-guard/staking-module-guard.service.ts
@@ -20,6 +20,11 @@ import { DeepReadonly } from 'common/ts-utils';
 import { Configuration } from 'common/config';
 import { getLegacyModuleWCs } from '../withdrawal-credentials';
 
+type UsedKeyExpectedWC = {
+  moduleAddress: string;
+  wc: string;
+};
+
 @Injectable()
 export class StakingModuleGuardService {
   constructor(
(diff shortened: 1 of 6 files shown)


F4ever (dismissed):

src/guardian/staking-module-guard/staking-module-guard.service.ts
```diff
@@ -105,26 +116,29 @@ export class StakingModuleGuardService {
...
       depositedEvents,
-      usedKeyExpectedWC,
+      usedKeyAcceptableWCs,
     );
     const lidoWCs = new Set(Object.values(moduleAddressToWC));
 
     const frontRunnedKeys: string[] = [];
     for (const [pubkey, event] of earliestDeposits) {
+      const acceptableWCs = usedKeyAcceptableWCs.get(pubkey);
+      // A legacy WC (e.g. the pre-Merge 0x00 BLS WC) is a valid lido WC even
```
avsetsin: The Merge (Bellatrix HF) didn't introduce 0x01 WCs type or withdrawals

eddort (dismissed):

src/guardian/staking-module-guard/staking-module-guard.service.ts
```diff
@@ -143,26 +157,28 @@ export class StakingModuleGuardService {
...
     for (const [pubkey, event] of earliestDeposits) {
-      const expectedWC = usedKeyExpectedWC.get(pubkey);
-      if (lidoWCs.has(event.wc) && event.wc !== expectedWC) {
+      const acceptableWCs = usedKeyAcceptableWCs.get(pubkey);
+      // Acceptable WCs (current + legacy) are fine; only a lido WC that is
+      // neither is a wrong-type deposit.
+      if (acceptableWCs?.has(event.wc)) continue;
+      if (lidoWCs.has(event.wc)) {
         wrongWCTypeKeys.push(pubkey);
       }
```
avsetsin: Consider ways to improve the code's readability; now, it takes some effort to understand what's going on here. Maybe smth like this:

```ts
const hasAcceptableWC = acceptableWCs.has(event.wc);
const hasOtherLidoWC = !hasAcceptableWC && lidoWCs.has(event.wc);

if (hasOtherLidoWC) {
  otherLidoWCsKeys.push(pubkey);
}
```

// note, that `wrongWCTypeKeys` name doesn't fully describe what's stored there. The check is not based on the type of WCs but on all bytes32. So something like `otherLidoWCsKeys` would be a better fit

src/guardian/staking-module-guard/staking-module-guard.service.ts
avsetsin, line 174: The method doesn't check the WCs type, it checks the entire WCs of other modules. This implicit knowledge of how another method works, the one that collects these WCs for each module (`fetchModuleWithdrawalCredentials`), could backfire at some point. Consider renaming

src/guardian/withdrawal-credentials.ts
```diff
@@ -21,3 +21,44 @@ export function buildModuleWc(
...
+  number,
+  Record<string, string[]>
+> = {
+  // mainnet, Curated module (module id 1)
+  1: {
+    '0x55032650b14df07b85bf18a3a3ec8e0af2e028d5': [
+      '0x009690e5d4472c7c0dbdf490425d89862535d2a52fb686333f3a0a9ff5d2125e',
+    ],
+  },
+};
```
avsetsin, line 59: Consider adding something like a `validUntilBlock`, after which such WCs are considered invalid. Otherwise, new deposits with such WCs will not trigger the pause

src/guardian/staking-module-guard/staking-module-guard.service.ts
eddort: fixed

src/guardian/withdrawal-credentials.ts
eddort: fixed

src/guardian/staking-module-guard/staking-module-guard.service.ts
eddort: fixed

src/guardian/staking-module-guard/staking-module-guard.service.ts
eddort: I agree. We plan to merge checkHistoricalFrontRun and checkWrongWCType after the release, since their logic is very close. This should resolve the naming issue as well.

F4ever (approved):

PR #335, opened 2026-07-23
feat: Execution Delegation Framework support

eddort:
This branch teaches the guardian daemon to work through the Execution
Delegation Framework. On DSM v5 the guardian identity is a
`DelegationContract`; the daemon signs with the delegate key that the
contract currently recognizes. DSM v4 keeps the legacy EOA flow. DSM v3
support is removed; the daemon fails closed below v4.

## Main changes

- Per-block guardian execution context: the daemon reads the DSM `VERSION()`
  and the contract's `getDelegate()` on every new block, then selects the
  matching configured key. The v4-to-v5 cutover and a delegate rotation need
  no restart. The mode is logged every block
  (`Guardian execution mode: edf | legacy-eoa`).
- Fail-closed identity rules: startup requires a valid
  `DELEGATION_CONTRACT_ADDRESS`; a terminated contract, a missing delegate, a
  missing ERC-1271 interface, a non-guardian contract, or an unmatched
  delegate key each stop the cycle with a specific error. There is no
  fallback to the EOA on v5.
- Key configuration is a pool of two equal keys: `WALLET_PRIVATE_KEY` and
  `WALLET_PRIVATE_KEY_2`, each with a `_FILE` variant. On DSM v4 the daemon
  signs with the first key; on DSM v5 it signs with whichever key matches the
  on-chain delegate. A rotation stages the next key in the free slot; after
  the switch the retired key is removed.
- All version-specific behavior lives in one strategy module
  (`src/contracts/security/dsm-version.strategy.ts`): digest layout,
  signature layout, message-bus event version, and the transaction path.
  On v5, `pauseDeposits` and `unvetSigningKeys` go through
  `DelegationContract.execute()`.
- Message bus: v5 events (`MessageDepositV2`, `MessagePauseV4`,
  `MessageUnvetV2`) publish the 65-byte `r||s||v` signature that DSM v5
  accepts; v4 events keep the EIP-2098 compact form. Both coexist on the bus.
- New ABIs (DSM v5, DelegationContract, DelegationFactory, Locator, proxy)
  and pinned bytecode fixtures from EDF `1a3a87d` for the e2e suite.

## Test coverage

- Unit tests: config validation of the new keys, wallet selection against the
  on-chain delegate, the strategy module, and the execution-context logging.
- A locator-transition e2e spec covers the live v4-to-v5 flip: legacy flow
  before, delegate flow after, rotation to the second key, and the
  fail-closed revoked state.
- Data-bus e2e specs cover delegate sender resolution and the v5 message
  round trip.


src/wallet/wallet.service.ts
```diff
@@ -216,14 +279,18 @@ export class WalletService implements OnModuleInit {
   public async signPauseDataV3({
```
F4ever: maybe add `signPauseDataV5` method and do DSM version check outside?

test/dsm-version-adapter.e2e-spec.ts
```diff
@@ -0,0 +1,133 @@
...
+    const config = {
+      DELEGATION_CONTRACT_ADDRESS: process.env.DELEGATION_CONTRACT_ADDRESS,
+    } as Configuration;
+    provider = new SimpleFallbackJsonRpcBatchProvider(
+      {
+        urls: [TEST_SERVER_URL],
+        network: CHAIN_ID,
+        maxRetries: 1,
+        logRetries: false,
+        logSuccessfulAttempts: false,
```
chasingrainbows, line 48: `logSuccessfulAttempts` does not exist on `SimpleFallbackProviderConfig` in `@lido-nestjs/execution@1.23.0` — the interface only defines `logRetries`, so `tsc --noEmit` fails with TS2345 here and at the same provider construction in `test/edf-locator-transition.e2e-spec.ts:394`.

Since the e2e suites run through `ts-jest` with full type-checking (`isolatedModules` is not set in `test/jest-e2e.json`), both new EDF specs fail to compile before any test executes — `yarn typecheck` and `yarn test:e2e` both break on this.

src/common/config/in-memory-configuration.ts
```diff
@@ -170,6 +209,9 @@ export class InMemoryConfiguration implements Configuration {
   @IsString()
   LOCATOR_DEVNET_ADDRESS = '';
 
+  @IsEthereumAddress()
```
chasingrainbows, line 212: `DELEGATION_CONTRACT_ADDRESS` has `@IsEthereumAddress()` without `@IsOptional()` or `@ValidateIf`, so config validation requires a valid address in every deployment. The runtime check in `SecurityService.getGuardianExecutionContext` needs this address only when DSM version is 5. As a result, a v4-only deployment does not boot without this variable, and operators have to set a placeholder address - validation cannot tell a placeholder from a real delegation contract. Validating it only when non-empty (like `WALLET_PRIVATE_KEY_2`) or only for the v5 path would match the runtime rule.

test/edf-locator-transition.e2e-spec.ts
```diff
@@ -0,0 +1,890 @@
...
+const ERC1271_INVALID_VALUE = '0xffffffff';
+interface SignedMessages {
+  deposit: { digest: string; signature: Signature };
+  pause: { digest: string; signature: Signature };
+  unvet: { digest: string; signature: Signature };
+}
+
+/**
+ * The signature layout is part of the event signature, so each layout hashes to
+ * its own topic and the two event sets coexist on the bus. v3/v4 publish the
```
chasingrainbows, line 65: This comment and two more below (line 87 "the compact pair on v3/v4", line 317 "DSM v3/v4 take") still refer to DSM v3, but this PR removes v3 support - the daemon fails closed below v4. Keeping v3 in the version references makes the supported set unclear; only v4 and v5 exist in this codebase now.

src/contracts/security/security.service.ts
```diff
@@ -88,17 +192,15 @@ export class SecurityService {
    * Returns the guardian index in the list
    */
   public async getGuardianIndex(blockTag?: BlockTag): Promise<number> {
```
chasingrainbows, line 194: `getGuardianIndex` and `getGuardianAddress` have no production callers left - `initialize()` and the block data collector call `getGuardianExecutionContext` directly, and the only remaining callers are the unit tests of these two methods. Each call also re-runs the full execution-context resolution (up to 4 RPC calls on v5: `getGuardians`, `getDelegate`, `isTerminated`, `supportsInterface`) to return a single field. If they are not kept for external compatibility, they look like leftovers from the refactor.

.nvmrc
```diff
@@ -0,0 +1 @@
+18
```
F4ever, line 1: Hm

chasingrainbows (changes requested):

chasingrainbows (dismissed):

chasingrainbows (dismissed):

F4ever (approved):

PR #338, opened 2026-08-13
feat: edf metrics

eddort:
### Summary

- Added the `council_daemon_guardian_info` metric. It exposes the current execution mode, DSM version and address, guardian address, active wallet, and all configured wallets.
- Added an EDF readiness check during Council initialization.
- The check validates that `DELEGATION_CONTRACT_ADDRESS` contains contract code, has a valid active delegate, and that the delegate matches one of the configured wallets.
- Council logs `EDF setup is ready` when the configuration is valid. It logs `EDF setup is not ready` with a clear reason when the check fails.
- The readiness check is diagnostic and does not throw.

### Grafana example

```promql
max by (activeWalletAddress, mode) (
  council_daemon_guardian_info{
    env=~"$env",
    source="$source",
    isActive="true"
  }
)
```

Use this legend for a Stat panel:

```text
{{activeWalletAddress}} · {{mode}}
```

Set **Text mode** to **Name**.


src/contracts/security/security.service.ts
```diff
@@ -50,7 +50,11 @@ export class SecurityService {
     private config: Configuration,
   ) {}
 
-  public async initialize(blockTag: BlockTag): Promise<void> {
+  public async initialize(
```
chasingrainbows: `initialize()` is not covered by tests: nothing checks that it calls `logEdfReadiness` and returns the context from `getGuardianExecutionContext`. If `return context` is removed later, metric population in `GuardianService.onModuleInit` breaks while all tests stay green. A small `describe('initialize')` block asserting the call and the resolved value would close this gap.

src/guardian/guardian.service.ts
```diff
@@ -86,13 +90,14 @@ export class GuardianService implements OnModuleInit {
         const stakingRouterModuleAddresses =
           await this.stakingRouterService.getStakingModulesAddresses(blockHash);
 
-        await Promise.all([
+        const [, guardianContext] = await Promise.all([
```
chasingrainbows: `onModuleInit` is not exercised in the spec, so the destructuring index and the `setGuardianExecutionContext(guardianContext)` call are untested. Today TypeScript would catch a reorder only because the other initializers return `void` - that protection is incidental. A test that mocks `securityService.initialize` with a known context and checks the gauge/`lastGuardianExecutionContext` would catch index regressions.

src/contracts/security/security.service.ts
```diff
@@ -60,6 +64,100 @@ export class SecurityService {
...
+        );
+        return;
+      }
+
+      this.logger.log('EDF setup is ready', {
+        delegationContractAddress,
+        delegateAddress,
+        configuredWalletAddresses,
+      });
+    } catch (error) {
```
chasingrainbows: The catch keeps only `error.message`, so an unexpected bug here (e.g. a TypeError from a bad response shape) produces a single warn line without a stack trace, worded like a routine config issue. Passing the original error object to `logger.warn` alongside the reason would keep the stack while staying non-throwing.

src/contracts/security/security.service.ts
```diff
@@ -60,6 +64,100 @@ export class SecurityService {
...
     }
+
+    return context;
+  }
+
+  /**
+   * Reports whether the configured delegation contract can be used after the
+   * DSM switches to v5. This diagnostic must never block legacy DSM startup.
+   */
+  public async logEdfReadiness(blockTag: BlockTag): Promise<void> {
```
chasingrainbows: The readiness check skips the `isTerminated()` and ERC-1271 checks that `getGuardianExecutionContext` performs on the v5 path, so "EDF setup is ready" can be logged for a contract the daemon will later reject. Fine as a coarse diagnostic - just worth keeping in mind if this log line is treated as an authoritative readiness signal.

chasingrainbows (approved):

